PR #2, opened 2021-11-19
Add properties to set stun and turn servers

thiblahute:
And dot consumer pipelines on state changes.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -30,6 +30,7 @@ static CAT: Lazy<gst::DebugCategory> = Lazy::new(|| {
 const RTP_TWCC_URI: &str =
     "http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01";
 
+const DEFAULT_STUN_SERVER: Option<&str> = Some("stun://stun.l.google.com:19302");
 const DEFAULT_CONGESTION_CONTROL: WebRTCSinkCongestionControl =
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,7 +99,7 @@ In the third, run:
 
 ``` shell
 export GST_PLUGIN_PATH=$PWD/target/debug:$GST_PLUGIN_PATH
-gst-launch-1.0 webrtcsink name=ws videotestsrc ! ws. audiotestsrc ! ws.
+gst-launch-1.0 webrtcsink name=ws stun-server=stun://stun.l.google.com:19302 videotestsrc ! ws. audiotestsrc ! ws.
```
MathieuDuponchelle, line 102: No need for a stun-server here, the instructions are explicitly to run the example locally :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -1239,6 +1243,18 @@ impl WebRTCSink {
             .set_property_from_str("bundle-policy", "max-bundle")
             .unwrap();
 
+        if let Some(stun_server) = self.settings.lock().unwrap().stun_server.as_ref() {
```
MathieuDuponchelle, line 1246: elsewhere locking order is settings -> state, let's keep it that way :)

MathieuDuponchelle:
lgtm otherwise, thanks :)

README.md
thiblahute: Well, "locally" is not necessarily so locally as you know ;)

PR #3, opened 2021-11-19
sink: Add a `new-webrtcbin` signal

thiblahute:
Allowing user to add a data channel for example

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -1379,7 +1379,7 @@ impl WebRTCSink {
 
         let mut consumer = Consumer {
             pipeline: pipeline.clone(),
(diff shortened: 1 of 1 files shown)


plugins/src/webrtcsink/imp.rs
```diff
@@ -1427,6 +1427,10 @@ impl WebRTCSink {
             }
         });
 
+        pipeline.set_state(gst::State::Ready)?;
+
+        element.emit_by_name("new-webrtcbin", &[&peer_id.to_value(), &webrtcbin.to_value()])?;
```
sdroege, line 1432: The `to_value()` calls are not needed here, and also you probably want to just `unwrap()` here. If this fails then you failed typing the signal name or used the wrong types/number of arguments.

PR #5, opened 2021-11-20
Update dependencies

sdroege:
This involves removing lots of unnecessary `unwrap()`s and simplifies
the `lookup_codecs()` function a bit.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -247,10 +247,10 @@ fn setup_encoding(
         .with_context(|| format!("Creating payloader {}", codec.payloader.name()))?;
     let parse_filter = make_element("capsfilter", None)?;
 
-    pay.set_property("pt", codec.payload as u32).unwrap();
+    pay.set_property("pt", codec.payload as u32);
 
     if let Some(ssrc) = ssrc {
-        pay.set_property("ssrc", ssrc).unwrap();
+        pay.set_property("ssrc", ssrc);
     }
 
     pipeline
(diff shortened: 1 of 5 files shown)


sdroege:
Depends on https://github.com/gtk-rs/gtk-rs-core/pull/386

plugins/src/webrtcsink/utils.rs
```diff
@@ -55,14 +55,10 @@ impl StreamProducer {
...
 
-        consumer.set_property("max-buffers", 0u64).unwrap();
-        consumer.set_property("max-bytes", 0u64).unwrap();
-        consumer
-            .set_property("max-time", 500 * gst::ClockTime::MSECOND)
-            .unwrap();
-        consumer
-            .set_property_from_str("leaky-type", "downstream")
-            .unwrap();
+        consumer.set_property("max-buffers", 0u64);
```
MathieuDuponchelle, line 58: I was wondering when I ported another element, what was the rationale for removing the checks in the bindings? The fact that everyone ended up unwrapping anyway? What happens now if a property is the wrong type / doesn't exist? Silently fails as `gobject_set` would? I must say I appreciated not having to look up exact types everytime, and having the thing panic at runtime if I hadn't guessed correctly :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -992,38 +961,34 @@ impl WebRTCSink {
...
+                        gst_warning!(CAT, obj: &self.instance(),
                                 "Too many formats for available payload type range, ignoring {}",
                                 s);
-                            None
-                        }
-                    })
-                    .flatten()
+                        None
+                    }
+                })
```
MathieuDuponchelle, line 991: nicer without the `flatten()``, tnx :)

MathieuDuponchelle:
tnx :)

plugins/src/webrtcsink/utils.rs
sdroege: It will directly panic if you typoed the property name or used the wrong type. `set_property()` is basically `try_set_property().unwrap()` now (and if you want the old behaviour there's `try_set_property()`).

The reason for this change was that everybody `unwrap`ed anyway and it's more a programmer error (in 99% of the cases) than an error you are supposed to handle.

MathieuDuponchelle: Ah I see, thanks a lot

PR #6, opened 2021-11-24
Add properties to set minimum and maximum bitrate values

thiblahute:
Instead of hardcoding those.

Also update dependencies again.


MathieuDuponchelle:
lgtm, thanks :)

PR #8, opened 2021-12-03
Allow changeing Congestion control algorithm at runtime

thiblahute:


MathieuDuponchelle:
Why?

thiblahute:
To demonstrate difference of behavior in real time, people like to see real results. Implementation is simple enough that it is worth it to me.

plugins/src/webrtcsink/imp.rs
```diff
@@ -2084,9 +2084,29 @@ impl ObjectImpl for WebRTCSink {
...
+                let new_heuristic = value
                     .get::<WebRTCSinkCongestionControl>()
                     .expect("type checked upstream");
+                if new_heuristic != settings.cc_heuristic {
+                    settings.cc_heuristic = new_heuristic;
+
+                    let mut state = self.state.lock().unwrap();
+                    for (peer_id, consumer) in state.consumers.iter_mut() {
+                        match new_heuristic {
+                            WebRTCSinkCongestionControl::Disabled => {
```
MathieuDuponchelle, line 2096: You want to set video encoder's bitrates to the max value in that case.

thiblahute: Ah, I forgot to cherry-pick https://github.com/centricular/webrtcsink/pull/8/commits/2d8f170261de768371933b30807fc539bd23f5d0

thiblahute: The behavior with that last patch is that we set the right bitrate right away.

MathieuDuponchelle: OK, but still you're not resetting the bitrate in the property

MathieuDuponchelle: merging for now, I'll fix locally

PR #9, opened 2021-12-23
nvh264enc: Do not wrongly set vbv-buffer-size and use the cbr-ld-hq r…

thiblahute:
…c-mode

vbv-buffer-size which is in kbits can't be set while playing so we
can't really use it for our use case.


MathieuDuponchelle:
hah, for some reason I thought we had already fixed that, thanks :)

PR #10, opened 2021-12-23
stat-frontend: Allow passing a URL to the stat producer

thiblahute:


plugins/examples/webrtcsink-stats/src/App.svelte
```diff
@@ -80,7 +80,14 @@
   }
 
   const fetchStats = () => {
-    ws = new WebSocket("ws://127.0.0.1:8484");
+    const urlParams = new URLSearchParams(window.location.search);
```
MathieuDuponchelle, line 83: can you document that behavior in https://github.com/centricular/webrtcsink/blob/main/plugins/examples/webrtcsink-stats/README.md?

PR #11, opened 2021-12-23
Add support for cuda and GL memory 

thiblahute:


plugins/src/webrtcsink/imp.rs
```diff
@@ -1865,17 +1913,23 @@ impl WebRTCSink {
...
-        src.link(&capsfilter)
-            .with_context(|| format!("Running discovery pipeline for caps {}", caps))?;
+        elements.push(make_converter_for_video_caps(caps)?);
 
-        let (_, _, pay) = setup_encoding(&pipe.0, &capsfilter, codec, None, true)?;
+        let capsfilter = make_element("capsfilter", None)?;
+        elements.push(capsfilter.clone());
+        let elements_slice = &elements.iter().collect::<Vec<_>>();
+        pipe.0.add_many(elements_slice).unwrap();
+        gst::Element::link_many(elements_slice).unwrap();
```
MathieuDuponchelle, line 1930: don't unwrap here, it shouldn't be fatal, instead keep the `.with_context(|| format!("Running discovery pipeline for caps {}", caps))?;`

MathieuDuponchelle:
lgtm otherwise, thanks :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -242,22 +245,45 @@ impl Default for State {
     }
 }
 
+fn make_converter_for_video_caps(caps: &gst::Caps) -> Result<gst::Element, Error> {
+    assert!(caps.size() == 1);
```
sdroege, line 249:
```suggestion
    assert_eq!(caps.size(), 1);
```

But maybe you actually want

```suggestion
    assert!(caps.is_fixed());
```

?

plugins/src/webrtcsink/imp.rs
```diff
@@ -242,22 +245,45 @@ impl Default for State {
...
+
+    for feature in caps.features(0) {
+        if feature.contains(CUDA_MEMORY_FEATURE) {
+            return Ok(gst::parse_bin_from_description(
+                "cudaupload ! cudaconvert ! cudascale ! videorate drop-only=true",
```
sdroege, line 254:
```suggestion
                "cudaupload ! cudaconvert ! cudascale ! videorate drop-only=true skip-to-first=true",
```

Is maybe wanted here. Was already wrong before then though :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -242,22 +245,45 @@ impl Default for State {
...
+                true,
+            )?
+            .upcast());
+        } else if feature.contains(GL_MEMORY_FEATURE) {
+            return Ok(gst::parse_bin_from_description(
+                "glupload ! glcolorconvert ! glcolorscale ! videorate drop-only=true",
```
sdroege, line 260:
```suggestion
                "glupload ! glcolorconvert ! glcolorscale ! videorate drop-only=true skip-to-first=true",
```

plugins/src/webrtcsink/imp.rs
```diff
@@ -242,22 +245,45 @@ impl Default for State {
...
+                true,
+            )?
+            .upcast());
+        }
+    }
+
+    Ok(gst::parse_bin_from_description(
+        "videoconvert ! videoscale ! videorate drop-only=true",
```
sdroege, line 268:
```suggestion
        "videoconvert ! videoscale ! videorate drop-only=true skip-to-first=true",
```

plugins/src/webrtcsink/imp.rs
```diff
@@ -302,19 +328,23 @@ fn setup_encoding(
...
+        let structure_builder = if codec.encoder.name() == "nvh264enc" {
+            // Quirk: nvh264enc can perform conversion from RGB formats, but
+            // doesn't advertise / negotiate colorimetry correctly, leading
+            // to incorrect color display in Chrome (but interestingly not in
+            // Firefox). In any case, restrict to exclude RGB formats altogether,
+            // and let videoconvert do the conversion properly if needed.
+            structure_builder.field("format", &gst::List::new(&[&"NV12", &"YV12", &"I420"]))
+        } else {
+            structure_builder
+        };
```
sdroege, line 344:
```suggestion
        let mut structure_builder = gst::Structure::builder("video/x-raw")
            .field("pixel-aspect-ratio", gst::Fraction::new(1, 1));

        if codec.encoder.name() == "nvh264enc" {
            // Quirk: nvh264enc can perform conversion from RGB formats, but
            // doesn't advertise / negotiate colorimetry correctly, leading
            // to incorrect color display in Chrome (but interestingly not in
            // Firefox). In any case, restrict to exclude RGB formats altogether,
            // and let videoconvert do the conversion properly if needed.
            structure_builder = structure_builder.field("format", &gst::List::new(&[&"NV12", &"YV12", &"I420"]));
        }
```

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: interesting, I haven't checked the code but I would have thought `drop-only` would imply that?

sdroege: That could make sense but I don't think it does that

MathieuDuponchelle: OK, can't hurt to add anyway, it matches the intention and I can't tell from a quick read of the source code whether `drop-only` alone results in the same behaviour

plugins/src/webrtcsink/imp.rs
thiblahute: Indeed :-)

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: Did you forget to update that one? :)

MathieuDuponchelle:
thanks :)

PR #13, opened 2021-12-24
Implement the GstNavigation interface using a dedicated DataChannel protocol

thiblahute:
This allows interacting with the source element from within the browser very easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,27 @@ for the list of properties.
 
 [the source code]: plugins/src/signaller/imp.rs
 
+
+### Enable 'navigation' a.k.a user interactivity with the content
+
+`webrtcsink` implements the [`GstNavigation`] interface which allows interacting
+with the content, for example move with your mouse, entering keys with the
+keyboard, etc... On top of that a `WebRTCDataChannel` based protocol has been
+implemented and can be activated with the `enable-data-channel-navigation=true`
+property. The [demo](www/) implements the protocol and you can easily test this
+feature, using the [`wpesrc`] for example.
+
+As an example, the following pipeline allows you to navigate the GStreamer
+documentation inside the video running within your web browser (in
+http://127.0.0.1:8000 if you followed previous steps of that readme):
+
+```
+gst-launch-1.0 wpesrc location=https://gstreamer.freedesktop.org/documentation/ ! webrtcsink enable-data-channel-navigation=true
+```
+
+[`GstNavigation`]: https://gstreamer.freedesktop.org/documentation/video/gstnavigation.html
+[`wpesrc`]: https://gstreamer.freedesktop.org/documentation/wpe/wpesrc.html
+
 ## Testing congestion control
 
 For the purpose of testing congestion in a reproducible manner, a
--- a/plugins/Cargo.toml
+++ b/plugins/Cargo.toml
@@ -11,7 +11,7 @@ build = "build.rs"
 [dependencies]
 gst = { package = "gstreamer", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20", "ser_de"] }
 gst-app = { package = "gstreamer-app", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20"] }
-gst-video = { package = "gstreamer-video", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20"] }
+gst-video = { package = "gstreamer-video", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20", "ser_de"] }
 gst-webrtc = { package = "gstreamer-webrtc", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20"] }
 gst-sdp = { package = "gstreamer-sdp", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20"] }
 gst-rtp = { package = "gstreamer-rtp", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20"] }
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -2,8 +2,11 @@ use anyhow::Context;
 use gst::glib;
 use gst::prelude::*;
 use gst::subclass::prelude::*;
+use gst_video::prelude::*;
+use gst_video::subclass::prelude::*;
 use gst::{gst_debug, gst_error, gst_info, gst_log, gst_trace, gst_warning};
 use gst_rtp::prelude::*;
+use gst_webrtc::WebRTCDataChannel;
 
 use async_std::task;
 use futures::prelude::*;
@@ -41,6 +44,7 @@ const DEFAULT_CONGESTION_CONTROL: WebRTCSinkCongestionControl =
     WebRTCSinkCongestionControl::Homegrown;
 const DEFAULT_DO_FEC: bool = true;
 const DEFAULT_DO_RETRANSMISSION: bool = true;
+const DEFAULT_ENABLE_DATA_CHANNEL_NAVIGATION: bool = false;
 
 /// User configuration
 struct Settings {
@@ -53,6 +57,7 @@ struct Settings {
     max_bitrate: u32,
     do_fec: bool,
     do_retransmission: bool,
+    enable_data_channel_navigation: bool,
 }
 
 /// Represents a codec we can offer
@@ -189,10 +194,29 @@ struct State {
     audio_serial: u32,
     video_serial: u32,
     streams: HashMap<String, InputStream>,
+    navigation_handler: Option<NavigationEventHandler>,
 }
 
+fn create_navigation_event<N: IsA< gst_video::Navigation>>(sink: &N, msg: &str) {
+    let event: Result<gst_video::NavigationEvent, _> = serde_json::from_str(msg);
+
+    if let Ok(event) = event {
+        sink.send_event(event.structure());
+    } else {
+        gst_error!(CAT, "Invalid navigation event: {:?}", msg);
+    }
+
+
+}
 /// Simple utility for tearing down a pipeline cleanly
 struct PipelineWrapper(gst::Pipeline);
+
+// Structure to generate GstNavigation event from a WebRTCDataChannel
+#[derive(Debug)]
+struct NavigationEventHandler {
+    channel: WebRTCDataChannel,
+    message_sig: glib::SignalHandlerId,
+}
 
 /// Our instance structure
 #[derive(Default)]
@@ -219,6 +243,7 @@ impl Default for Settings {
             max_bitrate: DEFAULT_MAX_BITRATE,
             do_fec: DEFAULT_DO_FEC,
             do_retransmission: DEFAULT_DO_RETRANSMISSION,
+            enable_data_channel_navigation: DEFAULT_ENABLE_DATA_CHANNEL_NAVIGATION,
         }
     }
 }
@@ -238,6 +263,7 @@ impl Default for State {
             audio_serial: 0,
             video_serial: 0,
             streams: HashMap::new(),
+            navigation_handler: None,
         }
     }
 }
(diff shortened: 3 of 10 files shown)


plugins/src/webrtcsink/imp.rs
```diff
@@ -189,11 +194,30 @@ struct State {
...
     video_serial: u32,
     streams: HashMap<String, InputStream>,
+    navigation_handler: Option<NavigationEventHandler>,
 }
 
+fn create_navigation_event(sink: gst_video::Navigation, msg: &str) {
+    let event: Result<gst_video::NavigationEvent, _> = serde_json::from_str(msg);
+
+    if let Ok(event) = event {
+        sink.send_event(event.build().structure().unwrap().to_owned());
```
sdroege, line 204: This seems a bit suboptimal API-wise. Why first build a `gst::Event` just to get a reference to the `gst::Structure` it contains and then creating a copy of that?

That probably should be improved a bit in the bindings. How is this usually done in C code?

plugins/src/webrtcsink/imp.rs
```diff
@@ -1134,6 +1160,31 @@ impl InputStream {
...
+        let channel = webrtcbin.emit_by_name::<WebRTCDataChannel>(
+            "create-data-channel",
+            &[&"input", &gst::Structure::new_empty("nothing")],
+        );
+
+        let weak_element = element.downgrade();
+        Self {
+            message_sig: channel.connect("on-message-string", false, move |values| {
+                if let Some(element) = weak_element.upgrade() {
+                    create_navigation_event(element.upcast(), values[1].get::<&str>().unwrap());
```
sdroege, line 1178:
```suggestion
                    let _channel = values[0].get::<WebRTCDataChannel>().unwrap();
                    let msg = values[1].get::<&str>().unwrap();
                    create_navigation_event(element.upcast(), msg);
```

I usually get all of the arguments in their proper types in the beginning, even the ones not actually used, to make sure the expectation of the signal signature is correct.

plugins/src/webrtcsink/imp.rs
```diff
@@ -189,11 +194,30 @@ struct State {
     audio_serial: u32,
     video_serial: u32,
     streams: HashMap<String, InputStream>,
+    navigation_handler: Option<NavigationEventHandler>,
 }
 
+fn create_navigation_event(sink: gst_video::Navigation, msg: &str) {
```
sdroege, line 200:
```suggestion
fn create_navigation_event<IsA<N: gst_video::Navigation>(sink: &N, msg: &str) {
```

or at least

```suggestion
fn create_navigation_event(sink: &gst_video::Navigation, msg: &str) {
```

The former should avoid the `upcast()` below, for the latter you'd need to do `upcast_ref()` instead. In either case, less destructive :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -1134,6 +1160,31 @@ impl InputStream {
...
 
+impl NavigationEventHandler {
+    pub fn new(
+        element: &super::WebRTCSink,
+        webrtcbin: &gst::Element,
+    ) -> Self {
+
+        let channel = webrtcbin.emit_by_name::<WebRTCDataChannel>(
+            "create-data-channel",
+            &[&"input", &gst::Structure::new_empty("nothing")],
```
sdroege, line 1171:
```suggestion
            &[&"input", &None::<gst::Structure>],
```

Do you have to actually pass an empty dummy structure here?

plugins/src/webrtcsink/imp.rs
```diff
@@ -2105,7 +2161,7 @@ impl ObjectSubclass for WebRTCSink {
     const NAME: &'static str = "RsWebRTCSink";
     type Type = super::WebRTCSink;
     type ParentType = gst::Bin;
-    type Interfaces = (gst::ChildProxy,);
+    type Interfaces = (gst::ChildProxy, gst_video::Navigation,);
```
sdroege, line 2164:
```suggestion
    type Interfaces = (gst::ChildProxy, gst_video::Navigation);
```

The trailing comma is only need for 1-tuples to distinguish them from a plain value

plugins/src/webrtcsink/imp.rs
```diff
@@ -2519,3 +2590,22 @@ impl ChildProxyImpl for WebRTCSink {
...
+        let mut state = self.state.lock().unwrap();
+
+        state
+            .streams
+            .iter_mut()
+            .for_each(|(_, stream)| {
+                if stream.sink_pad.name().starts_with("video_") {
+                    gst_log!(CAT, "Navigating to: {:?}", event_def);
+                    // FIXME: Handle multi tracks.
+                    if !stream.sink_pad.push_event(gst::event::Navigation::new(event_def.clone())) {
```
sdroege, line 2605: Why not

```rust
let event = gst::event::Navigation::new(event_def);
```

at the very top and then just

```rust
if !stream.sink_pad.push_event(event.clone());
```

in the loop? That way you push the same event to all pads (refcounting!) instead of creating a copy of the whole structure for every pad.

plugins/src/webrtcsink/imp.rs
thiblahute: You can't do it in C, you need to use the GstNavigationInterface API as there is no API to create GstNavigation events in C. Using the navigation interface API here would make everything uglier as I already got the structures ready. I could add a `NavigationEvent.structure()` method in the bindings if you prefer.

plugins/src/webrtcsink/imp.rs
thiblahute: That also implies we could create a useless event and webrtcsink doesn't seem to really handle several streams of a type yet.

plugins/src/webrtcsink/imp.rs
thiblahute: Here: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/935

plugins/src/webrtcsink/imp.rs
thiblahute: Added that new method here: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/935

README.md
```diff
@@ -135,6 +135,28 @@ for the list of properties.
 
 [the source code]: plugins/src/signaller/imp.rs
 
+
+### Enable 'navigation' a.k.a user interactivity with the content
+
+`webrtcsink` implements the [`GstNavigation`] interface which allows interacting
+with the content, for example move with your mouse, entering keys without the
```
MathieuDuponchelle, line 142: with*out* the keyboard? :D

README.md
```diff
@@ -135,6 +135,28 @@ for the list of properties.
...
+keyboard, etc... On top of that a `WebRTCDataChannel` based protocol has been
+implemented and can be activated with the `enable-data-channel-navigation=true`
+property. The demo located in `www/` of this repository also implements the
```
MathieuDuponchelle, line 145: you can actually link to the demo folder instead, eg:

```
the [demo](www/) implements ...
```

should work

README.md
```diff
@@ -135,6 +135,28 @@ for the list of properties.
...
+protocol and you can easily test this feature , using the [`wpesrc`] for
+example.
+
+As an example, the following pipeline allows you to navigate the GStreamer
+documentation inside the video running within your web browser (in
+http://127.0.0.1:8000 if you followed previous steps of that readme):
+
+```
+gst-launch-1.0 wpesrc location=https://gstreamer.freedesktop.org/documentation/ ! webrtcsink enable-data-channel-navigation=true
```
MathieuDuponchelle, line 154: hah, that is so cool, I need to try this :)

plugins/Cargo.toml
```diff
@@ -11,7 +11,7 @@ build = "build.rs"
 [dependencies]
 gst = { package = "gstreamer", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20", "ser_de"] }
 gst-app = { package = "gstreamer-app", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20"] }
-gst-video = { package = "gstreamer-video", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_20"] }
+gst-video = { package = "gstreamer-video", git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["v1_18", "v1_20", "ser_de"] }
```
MathieuDuponchelle, line 14: doesn't `v1_20` already imply `v1_18` ?

plugins/src/webrtcsink/imp.rs
```diff
@@ -189,11 +194,30 @@ struct State {
...
+    }
+
+
+}
 /// Simple utility for tearing down a pipeline cleanly
 struct PipelineWrapper(gst::Pipeline);
 
+// Structure to generate GstNavigation event from a WebRTCDataChannel
+#[derive(Debug)]
+pub(crate) struct NavigationEventHandler {
```
MathieuDuponchelle, line 216: Why is `pub(crate)` needed here?

plugins/src/webrtcsink/imp.rs
```diff
@@ -1134,6 +1160,33 @@ impl InputStream {
...
+        let channel = webrtcbin.emit_by_name::<WebRTCDataChannel>(
+            "create-data-channel",
+            &[&"input", &None::<gst::Structure>],
+        );
+
+        let weak_element = element.downgrade();
+        Self {
+            message_sig: channel.connect("on-message-string", false, move |values| {
+                if let Some(element) = weak_element.upgrade() {
+                    let _channel = values[0].get::<WebRTCDataChannel>().unwrap();
```
MathieuDuponchelle, line 1178: this line doesn't seem useful?

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: I haven't actually tested that, but what makes you say it won't handle it? For instance for congestion control I take into account the number of input video streams :)

plugins/Cargo.toml
sdroege: Yes

plugins/src/webrtcsink/imp.rs
sdroege: See https://github.com/centricular/webrtcsink/pull/13#discussion_r775219287

www/input.js
```diff
@@ -0,0 +1,486 @@
...
+    _wheel(event) {
+        let data = {
+            "event": "MouseScroll",
+            "x": this.x,
+            "y": this.y,
+            "delta_x": -event.deltaX,
+            "delta_y": -event.deltaY,
+        };
+
+        console.error(JSON.stringify(data))
```
MathieuDuponchelle, line 212: If you don't mind making sure those new files you check in don't spam the console more than needed that would be nice :)

www/webrtc.js
```diff
@@ -240,20 +245,36 @@ function Session(our_id, peer_id, closed_callback) {
...
-            let receive_channel = event.channel;
-            let buffer = [];
+            console.log(`Data channel created: ${event.channel.label}`);
+            this.data_channel = event.channel;
+
+            video_element = this.getVideoElement();
+            if (video_element) {
+                this.input = new Input(video_element, (data) => {
+                    if (this.data_channel) {
+                        console.error(`Navigation data: ${data}`);
```
MathieuDuponchelle, line 255: `console.error` probably a bit noisy :)

www/webrtc.js
```diff
@@ -240,20 +245,36 @@ function Session(our_id, peer_id, closed_callback) {
...
+                        this.data_channel.send(data);
```
MathieuDuponchelle, line 256: I can't really tell just looking at this, what format is that data?

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: hah, I see, I don't usually do that but I don't mind either way :)

plugins/src/webrtcsink/imp.rs
thiblahute: I have seen a few places where some more has to be done, can't recall what exactly. Not saying it is not handled at all but I could notice it wasn't properly tested :-)

README.md
thiblahute: oO :-)

www/input.js
thiblahute: Oops, forgot to remove some debug in there, sorry.

MathieuDuponchelle:
ooc, have you made changes to those two source files you bring in (`input.js` / `keyboard.js`)? If yes, I think ideally we should check them in as is, then keep the changes in a separate commit, no?

thiblahute:
> ooc, have you made changes to those two source files you bring in (`input.js` / `keyboard.js`)? If yes, I think ideally we should check them in as is, then keep the changes in a separate commit, no?

For keyboard.js I adapted https://cdn.jsdelivr.net/npm/apache-guacamole-client@0.9.14/dist/guacamole.js and the diff is plain useless

```
✗ git diff --stat HEAD^ -- keyboard.js |cat
 www/keyboard.js | 15497 +++++++++++-------------------------------------------
 1 file changed, 2996 insertions(+), 12501 deletions(-)
 ```

so I am not sure if I should do that - what do you think?

For input.js it does make sense.

PR #14, opened 2021-12-26
Fix various clippy warnings

sdroege:


plugins/src/webrtcsink/imp.rs
```diff
@@ -85,6 +85,8 @@ struct WebRTCPad {
     /// The (fixed) caps of the corresponding input stream
     in_caps: gst::Caps,
     /// Our offer caps
+    // FIXME: Use this at some point
+    #[allow(dead_code)]
     caps: gst::Caps,
```
sdroege, line 90: Is this intentionally unused currently?

plugins/src/webrtcsink/imp.rs
```diff
@@ -461,7 +463,7 @@ impl VideoEncoder {
 
         let width = height.mul_div_ceil(ratio.numer(), ratio.denom()).unwrap();
 
-        width + 1 >> 1 << 1
+        (width + 1) & !1
```
sdroege, line 466: All the shifting you did was a bit confusing and clippy complained that some parenthesis should be added to make it less confusing, i.e. `((width + 1) >> 1) << 1`. But IMHO with the binary `&` it's clearer what happens here.

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: No, I think this was unused since the initial commit, it can be removed altogether

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: works for me, though tbh it took me a second to understand what `!1` did :)

MathieuDuponchelle:
lgtm thanks, tbh I'm quite happy I didn't have more clippy errors than that :)

MathieuDuponchelle:
@sdroege I don't know if setting up github actions here is worth it, given my plan would be to move this to `gst-plugins-rs` sooner rather than later

plugins/src/webrtcsink/imp.rs
sdroege: Ok, removed that one

MathieuDuponchelle:
thanks :)

PR #18, opened 2022-01-25
Update to gstreamer-rs 0.18 instead of using it from git

sdroege:
And update all other dependencies.


MathieuDuponchelle:
thanks!

PR #22, opened 2022-02-07
Add API to allow better tracking consumers

thiblahute:
See commit messages.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -856,7 +856,12 @@ impl CongestionController {
 }
 
 impl State {
-    fn finalize_consumer(&mut self, element: &super::WebRTCSink, consumer: Consumer, signal: bool) {
+    fn finalize_consumer(
+        &mut self,
+        element: &super::WebRTCSink,
+        consumer: &Consumer,
+        signal: bool,
+    ) {
         consumer.pipeline.debug_to_dot_file_with_ts(
             gst::DebugGraphDetails::all(),
             format!("removing-peer-{}-", consumer.peer_id,),
@@ -881,9 +886,12 @@ impl State {
         }
     }
 
-    fn remove_consumer(&mut self, element: &super::WebRTCSink, peer_id: &str, signal: bool) {
+    fn remove_consumer(&mut self, element: &super::WebRTCSink, peer_id: &str, signal: bool) -> Option<Consumer> {
         if let Some(consumer) = self.consumers.remove(peer_id) {
-            self.finalize_consumer(element, consumer, signal);
+            self.finalize_consumer(element, &consumer, signal);
+            Some(consumer)
+        } else {
+            None
         }
     }
 
@@ -1748,7 +1756,10 @@ impl WebRTCSink {
             return Err(WebRTCSinkError::NoConsumerWithId(peer_id.to_string()));
         }
 
-        state.remove_consumer(element, peer_id, signal);
+        if let Some(consumer) = state.remove_consumer(element, peer_id, signal) {
+            drop(state);
+            element.emit_by_name::<()>("consumer-removed", &[&peer_id, &consumer.webrtcbin]);
+        }
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


plugins/src/webrtcsink/imp.rs
```diff
@@ -1716,7 +1719,7 @@ impl WebRTCSink {
             }
         })?;
 
-        element.emit_by_name::<()>("new-webrtcbin", &[&peer_id, &webrtcbin]);
+        element.emit_by_name::<()>("consumer-added", &[&peer_id, &webrtcbin]);
```
MathieuDuponchelle, line 1722: so, I see you refactored the consumer removal code path to emit the signal without the lock, should we move that code to the end of the function and add a `drop (state)` before?

plugins/src/webrtcsink/imp.rs
```diff
@@ -2457,22 +2463,60 @@ impl ObjectImpl for WebRTCSink {
...
+                    glib::types::Type::UNIT.into(),
+                )
+                .build(),
+
+                /*
+                 * RsWebRTCSink::consumer_removed:
+                 * @consumer_id: Identifier of the consumer that was removed
+                 * @webrtcbin: The webrtcbin connected to the newly removed consumer
+                 *
+                 * This signal is emitted right after the connection from a consumer
```
MathieuDuponchelle, line 2488: connection with?

MathieuDuponchelle:
lgtm otherwise, tnx :)

plugins/src/webrtcsink/imp.rs
thiblahute: Good catch! Fixed.

thiblahute:
Also passed the code to `cargo fmt` as I introduced some "regressions" implementing the navigation interface.

plugins/src/webrtcsink/imp.rs
```diff
@@ -362,7 +360,8 @@ fn setup_encoding(
             // to incorrect color display in Chrome (but interestingly not in
             // Firefox). In any case, restrict to exclude RGB formats altogether,
             // and let videoconvert do the conversion properly if needed.
-            structure_builder = structure_builder.field("format", &gst::List::new(&[&"NV12", &"YV12", &"I420"]));
+            structure_builder =
+                structure_builder.field("format", &gst::List::new(&[&"NV12", &"YV12", &"I420"]));
```
sdroege, line 364: Why all the `&` :sob:
They'are all not needed and just add line-noise

plugins/src/webrtcsink/imp.rs
```diff
@@ -2457,22 +2469,67 @@ impl ObjectImpl for WebRTCSink {
         static SIGNALS: Lazy<Vec<glib::subclass::Signal>> = Lazy::new(|| {
             vec![
                 /*
-                 * RsWebRTCSink::new-webrtcbin:
-                 * @peer_id: Identifier of the peer associated with the consumer added
+                 * RsWebRTCSink::consumer-added:
```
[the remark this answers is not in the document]
MathieuDuponchelle: Not that I know of, wishful placeholders as far as I understand :)

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: not @thiblahute's fault, probably just me writing this code as it once needed to be written?

sdroege: Yeah, that was needed until 2-3 months ago

plugins/src/webrtcsink/imp.rs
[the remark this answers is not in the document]
thiblahute: Getting hotdoc to generate doc for rust elements should be pretty straight forward.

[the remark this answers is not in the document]
MathieuDuponchelle: depends on whether a new comment parser needs to be written, but yes if we're satisfied with gtk-doc syntax it should be a simple matter of running the scanner and build system plumbing basically

PR #23, opened 2022-02-09
change x264enc default bitrate to match nvh264enc (fixes #21)

### LINKED ISSUE #21: Is the default rate for x264enc correct?
https://github.com/centricular/webrtcsink/blob/main/plugins/src/webrtcsink/imp.rs#L393

Is that really supposed to be `25608u32` ?

machenmusik:
Assume default bitrate should be the same for a given codec (here, H264 is 2048u32) regardless of which encoder is used (x264enc, nvh264enc).


MathieuDuponchelle:
tnx, PR welcome for `start-bitrate` too in case you feel like it :)

PR #24, opened 2022-02-11
Do not reset caps when they do not change

thiblahute:
Otherwise a useless renegotiation will be triggered by capsfilter

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -538,12 +538,8 @@ impl VideoEncoder {
             _ => unreachable!(),
         }
 
-        let mut s = self
-            .filter
-            .property::<gst::Caps>("caps")
-            .structure(0)
-            .unwrap()
-            .to_owned();
+        let current_caps = self.filter.property::<gst::Caps>("caps");
+        let mut s = current_caps.structure(0).unwrap().to_owned();
 
         // Hardcoded thresholds, may be tuned further in the future, and
         // adapted according to the codec in use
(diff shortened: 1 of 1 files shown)


plugins/src/webrtcsink/imp.rs
```diff
@@ -524,9 +524,8 @@ impl VideoEncoder {
             _ => unreachable!(),
         }
 
-        let mut s = self
-            .filter
-            .property::<gst::Caps>("caps")
+        let current_caps = self.filter.property::<gst::Caps>("caps");
+        let mut s = current_caps.clone()
```
sdroege, line 528:
```suggestion
        let current_caps = self.filter.property::<gst::Caps>("caps");
        let mut s = current_caps
```

That cloning shouldn't be necessary as you're already getting a copy of the structure anyway. This probably even gives a clippy warning or not?

MathieuDuponchelle: ping @thiblahute :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -538,12 +538,8 @@ impl VideoEncoder {
...
         }
 
-        let mut s = self
-            .filter
-            .property::<gst::Caps>("caps")
-            .structure(0)
-            .unwrap()
-            .to_owned();
+        let current_caps = self.filter.property::<gst::Caps>("caps");
+        let mut s = current_caps.structure(0).unwrap().to_owned();
```
sdroege, line 542: You're optimistic here. The caps could be `None` or empty/any, in either case this would panic :)

MathieuDuponchelle: yes, we're OK here, also the code this replaces was doing that too :)

MathieuDuponchelle:
thanks :)

Issue #25, opened 2022-02-20
Expose datachannel messages more generally?


machenmusik:
The GstNavigation interface implemented is a bit archaic from the days of DVD navigation menus; as such, other methods of exposing the data channel functionality might be more useful?

For example, data channel messages could be made available as signals, or perhaps something like appsink could receive just those...

Thoughts welcome.

MathieuDuponchelle:
Hey, the idea behind that "default" data channel is for it to only transmit events that can be directly transformed into GStreamer events and forwarded upstream for interpretation / handling. For what it's worth I am going to extend it soon for another project to support sending other types of events (I need seek events), and also to allow specifying the transceiver the event applies to (useful in the case of multiple bundled video streams).

The navigation interface was indeed initially developed for DVD navigation, but it is still the standard mechanism GStreamer elements must implement for user input (eg `wpesrc` implements it).

If you need a custom protocol instead, you can simply add your own data channel (the consumer-added signal will give you access to `webrtcbin` at the right moment), and work from there :)

machenmusik:
@MathieuDuponchelle I tried this, but doing `create-data-channel` in response to `consumer-added` does something bad... after a few seconds, the client seems to re-request either audio or video, which fails and then things are torn down.

MathieuDuponchelle:
@machenmusik not sure what your exact problem is there, but I know that @thiblahute creates his own data channel with no issues with that method at least, so it *should work* if that helps with anything :)

machenmusik:
Thanks - what I discovered is that things seem to be happy if `create-data-channel` happens *before* the pipeline starts playing, but `consumer-added` only occurs *after* the pipeline is playing, and that seems to create a race condition that is easily reproducible with localhost connections.  Moving `consumer-added` to just before pipeline starts playing eliminates the issue even with localhost connections.

PR #26, opened 2022-02-23
move consumer-added before pipeline starts playing

machenmusik:
As described in https://github.com/centricular/webrtcsink/issues/25#issuecomment-1048465601

Thanks - what I discovered is that things seem to be happy if `create-data-channel` happens before the pipeline starts playing, but `consumer-added` only occurs after the pipeline is playing, and that seems to create a race condition that is easily reproducible with localhost connections. Moving `consumer-added` to just before pipeline starts playing eliminates the issue even with localhost connections.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -1728,6 +1728,8 @@ impl WebRTCSink {
             state.navigation_handler = Some(NavigationEventHandler::new(&element, &webrtcbin));
         }
 
+        element.emit_by_name::<()>("consumer-added", &[&peer_id, &webrtcbin]);
+
         pipeline.set_state(gst::State::Playing).map_err(|err| {
             WebRTCSinkError::ConsumerPipelineError {
                 peer_id: peer_id.to_string(),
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
oof yes, it should be called when the element is in state Ready, that was originally the case but I think we might have broken this recently, lemme check

MathieuDuponchelle:
Yes, my fault: https://github.com/centricular/webrtcsink/pull/22#pullrequestreview-875408738

MathieuDuponchelle:
My comment there still stands though, I'd rather the signal is emitted with no lock held. I think we want to add the consumer to state.consumers, drop(state), then emit the signal and finally set the state to Playing :)

machenmusik:
feel free to close and adjust your own version to your liking :-)

MathieuDuponchelle:
> feel free to close and adjust your own version to your liking - or edit this one :-)

Hah, well you look like you're on top of it, I figured the requested change was trivial enough that you'd just update your merge request :)

machenmusik:
FYI - tried your suggested reordering and it actually broke my use case (had to revert), so personally leaving this one as-is...

MathieuDuponchelle:
@machenmusik just to make sure then, can you try https://github.com/centricular/webrtcsink/pull/27 ? If that matched what you had done, can you report how it breaks ?

PR #27, opened 2022-02-23
webrtcsink: emit consumer-added with webrtcbin in the Ready state

MathieuDuponchelle:
In order for applications to potentially create data channels at
the correct moment

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -1360,11 +1360,11 @@ impl WebRTCSink {
         &self,
         element: &super::WebRTCSink,
         offer: gst_webrtc::WebRTCSessionDescription,
-        peer_id: String,
+        peer_id: &str,
     ) {
         let mut state = self.state.lock().unwrap();
 
-        if let Some(consumer) = state.consumers.get(&peer_id) {
+        if let Some(consumer) = state.consumers.get(peer_id) {
             consumer
                 .webrtcbin
                 .emit_by_name::<()>("set-local-description", &[&offer, &None::<gst::Promise>]);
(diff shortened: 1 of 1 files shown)


thiblahute:
lgtm, I moved it in https://github.com/centricular/webrtcsink/commit/22205c235d75575b40c74d00536bb82b3a1316c1 as you asked for it but I forgot about that restriction

machenmusik:
So that is exactly what I tried as per last discussion on #25, but it broke my use case, so I reverted to what is in #26.

MathieuDuponchelle:
OK, doesn't really make sense to me tbh, @thiblahute care to give it a spin with your application?

MathieuDuponchelle:
Anyway, probably a race condition, can you provide a test case by any chance @machenmusik ?

machenmusik:
what I wonder is whether the drop() call somehow invalidates the webrtcbin passed in `consumer-added` such that `create-data-channel` has unexpected side effects; otherwise, idk.

MathieuDuponchelle:
@machenmusik OK, can you gather logs for the failure / success cases ? (`GST_DEBUG=*webrtc*:6`)

machenmusik:
no unfortunately, but this code insertion at line 136 of plugins/examples/webrtcsink-stats-server.rs appears to be enough to trigger the issue as per https://github.com/centricular/webrtcsink/pull/27#issuecomment-1049329274
```
    ws.connect("consumer-added", false, move |values| {
        // get the webrtcbin reference
        let peer_id = values[1].get::<String>().unwrap(); //expect("Invalid peer_id");
        let webrtcbin = values[2].get::<gst::Element>().unwrap(); //expect("Invalid webrtcbin");

        info!("consumer-added, create-data-channel");
        let channelmaybe = webrtcbin.try_emit_by_name::<gst_webrtc::WebRTCDataChannel>(
            "create-data-channel",
            &[&"input", &None::<gst::Structure>],
        );
        if let Err(err) = channelmaybe {
            info!("create-data-channel error {}", err);
            return None
        }

        task::spawn(async move {
            let channel = channelmaybe.unwrap();

            info!("create-data-channel done");

            let mut interval = async_std::stream::interval(std::time::Duration::from_millis(1000));

            while interval.next().await.is_some() {
                info!("... tick");
            }
        });

        None
    });
```

machenmusik:
np!  if you can reproduce as well, I'd suggest (potentially adding something like that to the committed example / docs so we remember and) merging #25 and closing this one -- then addressing with later changes if you see fit

MathieuDuponchelle:
Just a note, I reproduce, investigating :)

MathieuDuponchelle:
Fixed, see the second commit message if interested :) Thanks a lot for sticking around to debug this, I will close #26 now.

machenmusik:
Makes sense.  Nice find!

PR #28, opened 2022-03-10
signalling: update protocol, reimplement in rust

MathieuDuponchelle:


www/webrtc.js
```diff
@@ -222,7 +232,10 @@ function Session(our_id, peer_id, closed_callback) {
     this.connectPeer = function() {
         this.setStatus("Connecting " + this.peer_id);
 
-        this.ws_conn.send("START_SESSION " + this.peer_id);
+        this.ws_conn.send(JSON.stringify({
+            "type": "startsession",
+            "peer_id": this.peer_id
```
sdroege, line 237: underscore here ...

www/webrtc.js
```diff
@@ -211,7 +218,10 @@ function Session(our_id, peer_id, closed_callback) {
         this.ws_conn = new WebSocket(ws_url);
         /* When connected, immediately register with the server */
         this.ws_conn.addEventListener('open', (event) => {
-            this.ws_conn.send('REGISTER CONSUMER');
+            this.ws_conn.send(JSON.stringify({
+                "type": "register",
+                "peer-type": "consumer"
```
sdroege, line 223: ... but dash here

plugins/src/signaller/imp.rs
```diff
@@ -28,17 +28,51 @@ struct State {
     receive_task_handle: Option<task::JoinHandle<()>>,
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
```
sdroege, line 31: Might want to move the message structs into a separate module

signalling/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+license = "MIT"
+description = "GStreamer WebRTC sink signalling server"
+repository = "https://github.com/centricular/webrtcsink/"
+
+[dependencies]
+anyhow = "1"
+async-std = { version = "1", features = ["unstable"] }
+async-tungstenite = { version = "0.16", features = ["async-std-runtime", "async-native-tls"] }
+serde = "1"
+serde_derive = "1"
```
sdroege, line 15:
```suggestion
serde = { version = "1", features = ["derive"] }
```

everywhere

signalling/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+serde_json = "1"
+clap = { version = "3", features = ["derive"] }
+tracing = { version = "0.1", features = ["log"] }
+tracing-subscriber = { version = "0.3", features = ["registry", "env-filter"] }
+tracing-log = "0.1"
+futures = "0.3"
+uuid = { version = "0.8", features = ["v4"] }
```
sdroege, line 22:
```suggestion
uuid = { version = "0.8", features = ["v4", "serde"] }
```

or not?

signalling/src/protocol/mod.rs
```diff
@@ -0,0 +1,126 @@
+/// The default protocol used by the signalling server
+use serde_derive::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(tag = "peer-type")]
+#[serde(rename_all = "lowercase")]
+/// Confirms registration
+pub enum RegisteredMessage {
```
sdroege, line 8: You're duplicating these definitions now or not? Best to put it into a single place :)

signalling/src/server/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+                sender: websocket_sender,
+            },
+        );
+    }
+
+    /// Run the server
+    pub async fn run(&self, host: &str, port: u16) -> Result<(), SignallingServerError> {
+        let addr = format!("{}:{}", host, port);
+
+        // Create the event loop and TCP listener we'll accept connections on.
```
sdroege, line 215: Should probably get some TLS support

signalling/src/server/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+        let try_socket = TcpListener::bind(&addr).await;
+        let listener = try_socket.expect("Failed to bind");
+        info!("Listening on: {}", addr);
+
+        let state_clone = self.state.clone();
+        while let Ok((stream, _)) = listener.accept().await {
+            task::spawn(Self::accept_connection(state_clone.clone(), stream));
```
sdroege, line 222: It might make sense to also make `Self::accept_connection` public. Then the application can do its connection setup itself and in the end pass the connection here for being taken care of further.

signalling/src/handlers/mod.rs
```diff
@@ -0,0 +1,1356 @@
...
+    /// with other peers
+    fn handle_message(
+        &mut self,
+        sender: &mut dyn MessageSender,
+        message: String,
+        peer_id: &str,
+    ) -> Result<(), Box<dyn std::error::Error>>;
+
+    /// A connection was closed, remove the peer and clean up
+    fn remove_peer(&mut self, sender: &mut dyn MessageSender, peer_id: &str);
```
sdroege, line 17: Should there maybe be another function for when a new peer appears?

signalling/src/server/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum SignallingServerError {}
+
+impl SignallingServer {
+    /// Instantiate the signalling server with a MessageHandler
+    /// implementation. Use DefaultMessageHandler for the default
+    /// protocol, or implement your own handler
+    pub fn new(message_handler: Box<dyn MessageHandler>) -> Self {
```
sdroege, line 57:
```suggestion
    pub fn new<H: MessageHandler>(message_handler: H) -> Self {
        let message_handler = Box::new(message_handler);
```

That's a bit nicer to use

signalling/src/server/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+                })
+                .unwrap(),
+                peer_id,
+            );
+        }
+
+        state.message_sender = Some(message_sender);
+    }
+
+    async fn accept_connection(state: Arc<Mutex<State>>, stream: TcpStream) {
```
sdroege, line 120: Should this maybe take an `AsyncRead + AsyncWrite` to be more generic? Otherwise you need to change things here depending on whether TLS is used or not

sdroege:
Generally looks fine to me

signalling/Cargo.toml
MathieuDuponchelle: No, peer ids are just required to be String

plugins/src/signaller/imp.rs
MathieuDuponchelle: done

protocol/src/lib.rs
```diff
@@ -0,0 +1,126 @@
+/// The default protocol used by the signalling server
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(tag = "peer-type")]
+#[serde(rename_all = "lowercase")]
```
sdroege, line 6: `peer-type` with dash but `lowercase` would cause other things to have an underscore

protocol/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// Messages directly forwarded from one peer to another
+    Peer(PeerMessage),
+    /// Provides the current list of consumer peers
+    List { producers: Vec<String> },
+    /// Notifies that an error occurred with the peer's current session
+    Error { details: String },
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "peer_type")]
```
sdroege, line 41: Here `peer_type` with an underscore

protocol/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// Notifies listeners that a producer was removed
+    ProducerRemoved { peer_id: String },
+    /// Instructs a peer to generate an offer
+    StartSession { peer_id: String },
+    /// Signals that the session the peer was in was ended
+    EndSession { peer_id: String },
+    /// Messages directly forwarded from one peer to another
+    Peer(PeerMessage),
+    /// Provides the current list of consumer peers
+    List { producers: Vec<String> },
```
sdroege, line 35:
```suggestion
    List { #[serde(default)] producers: Vec<String> },
```

Maybe

protocol/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "lowercase")]
+/// Contents of the peer message
+pub enum PeerMessageInner {
+    /// Conveys an ICE candidate
+    Ice {
+        /// The candidate string
+        candidate: String,
+        #[serde(rename = "sdpMLineIndex")]
```
sdroege, line 86: All other things are lowercase, this is camelcase. That doesn't seem ideal

protocol/src/lib.rs
sdroege: You probably want `snake_case` or `kebab-case` instead of `lowercase` and then make it all consistent everywhere.

protocol/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "lowercase")]
+/// Messages directly forwarded from one peer to another
+pub struct PeerMessage {
+    /// The identifier of the peer, which must be in a session with the sender
+    pub peer_id: String,
+    /// The contents of the message
+    pub peer_message: PeerMessageInner,
```
sdroege, line 100:
```suggestion
    #[serde(flatten)] pub peer_message: PeerMessageInner,
```

maybe?

plugins/src/signaller/imp.rs
```diff
@@ -93,7 +65,10 @@ impl Signaller {
         let (mut ws_sink, mut ws_stream) = ws.split();
 
         ws_sink
-            .send(WsMessage::Text("REGISTER PRODUCER".to_string()))
+            .send(WsMessage::Text(
+                serde_json::to_string(&p::IncomingMessage::Register(p::RegisterMessage::Producer))
```
sdroege, line 69: Maybe somewhere at the very top you want to wrap a [`with(|m| serde_json::to_string(m).unwrap())`](https://docs.rs/futures/latest/futures/sink/trait.SinkExt.html#method.with) and here to not repeat yourself. Same for the stream

signalling/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+authors = ["Mathieu Duponchelle <mathieu@centricular.com>"]
+license = "MIT"
+description = "GStreamer WebRTC sink signalling server"
+repository = "https://github.com/centricular/webrtcsink/"
+
+[dependencies]
+anyhow = "1"
+async-std = { version = "1", features = ["unstable", "attributes"] }
+async-native-tls = "0.3"
+async-tungstenite = { version = "0.16", features = ["async-std-runtime", "async-native-tls"] }
```
sdroege, line 14:
```suggestion
async-tungstenite = { version = "0.17", features = ["async-std-runtime", "async-native-tls"] }
```

you're outdated ;)

plugins/Cargo.toml
```diff
@@ -23,9 +23,9 @@ futures = "0.3"
 async-std = { version = "1", features = ["unstable"] }
 async-tungstenite = { version = "0.16", features = ["async-std-runtime", "async-native-tls"] }
```
sdroege, line 24:
```suggestion
async-tungstenite = { version = "0.17", features = ["async-std-runtime", "async-native-tls"] }
```

signalling/src/handlers/mod.rs
```diff
@@ -0,0 +1,1081 @@
...
+    }
+}
+
+impl Handler {
+    #[instrument(level = "debug", skip(stream))]
+    /// Create a handler. When extending the protocol, pass in an
+    /// always pending stream, then call handle() and consume items until
+    /// pending
+    pub fn new(
+        stream: Pin<Box<dyn Stream<Item = (String, Option<p::IncomingMessage>)> + Send>>,
```
sdroege, line 29: Maybe you want to add a `Default` impl that uses a `Pending` stream?

signalling/src/server/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+            ws_sink.close().await?;
+
+            Ok::<(), Error>(())
+        });
+
+        let mut tx = self.state.lock().unwrap().tx.clone();
+        let this_id_clone = this_id.clone();
+        let state_clone = self.state.clone();
+        let receive_task_handle = task::spawn(async move {
+            while let Some(msg) = async_std::stream::StreamExt::next(&mut ws_stream).await {
```
sdroege, line 137:
```suggestion
            while let Some(msg) = ws_stream.next().await {
```

?

signalling/src/handlers/mod.rs
```diff
@@ -0,0 +1,1081 @@
...
+        consumers: HashMap<String, Option<String>>,
+        listeners: HashSet<String>,
+    }
+}
+
+impl Handler {
+    #[instrument(level = "debug", skip(stream))]
+    /// Create a handler. When extending the protocol, pass in an
+    /// always pending stream, then call handle() and consume items until
+    /// pending
```
sdroege, line 27: I think technically more correct here would be not just a `Pending` stream, but you would also remember the `Waker` from the `Context` in `poll_next()` if there is currently no item available and then wake up the `Waker` whenever an outgoing message becomes available (i.e. `handle()` was called and produced something).

signalling/src/handlers/mod.rs
```diff
@@ -0,0 +1,1081 @@
...
+
+pin_project! {
+    #[must_use = "streams do nothing unless polled"]
+    pub struct Handler {
+        #[pin]
+        stream: Pin<Box<dyn Stream<Item=(String, Option<p::IncomingMessage>)> + Send>>,
+        items: VecDeque<(String, p::OutgoingMessage)>,
+        producers: HashMap<String, HashSet<String>>,
+        consumers: HashMap<String, Option<String>>,
+        listeners: HashSet<String>,
```
sdroege, line 19: I think you at least want to add a `pub type PeerId = String` so all these funny strings everywhere have some more meaning.

protocol/src/lib.rs
MathieuDuponchelle: ended up going camelCase for consistency `event.candidate.toJSON()` on the javascript side

protocol/src/lib.rs
MathieuDuponchelle: hrm, what's the point of that? This is a message we emit fwiw

protocol/src/lib.rs
MathieuDuponchelle: everything camelCase now

protocol/src/lib.rs
sdroege: Allowing to deserialize a message that doesn't contain the producers field at all

protocol/src/lib.rs
sdroege: Reason here is that it's probably nicer to read JSON

signalling/Cargo.toml
MathieuDuponchelle: updated, fortunately no changes needed

signalling/src/handlers/mod.rs
MathieuDuponchelle: perhaps in the future, but for now I have unpubbed handle() and removed mentions of it in the comment

signalling/src/handlers/mod.rs
MathieuDuponchelle: those fields are not public, but the type alias is nicer yes, done (without pub)

protocol/src/lib.rs
MathieuDuponchelle: hrm, the server will always provide it :)

signalling/src/handlers/mod.rs
MathieuDuponchelle: correct, handle was unpubbed however so obsolete (at least for now)

PR #31, opened 2022-03-23
protocol: allow registering with a display name

MathieuDuponchelle:
While the signalling server will allocate unique identifiers for
each peer, users may want to also pass their own identifiers.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -37,6 +37,7 @@ const RTP_TWCC_URI: &str =
     "http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01";
 
 const DEFAULT_STUN_SERVER: Option<&str> = Some("stun://stun.l.google.com:19302");
+const DEFAULT_DISPLAY_NAME: Option<&str> = None;
 const DEFAULT_MIN_BITRATE: u32 = 1000;
 
 /* I have found higher values to cause packet loss *somewhere* in
@@ -61,6 +62,7 @@ struct Settings {
     do_fec: bool,
     do_retransmission: bool,
     enable_data_channel_navigation: bool,
+    display_name: Option<String>,
 }
 
 /// Represents a codec we can offer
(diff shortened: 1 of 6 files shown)


plugins/src/webrtcsink/imp.rs
```diff
@@ -213,8 +215,8 @@ struct PipelineWrapper(gst::Pipeline);
 // Structure to generate GstNavigation event from a WebRTCDataChannel
 #[derive(Debug)]
 struct NavigationEventHandler {
-    channel: WebRTCDataChannel,
-    message_sig: glib::SignalHandlerId,
+    _channel: WebRTCDataChannel,
+    _message_sig: glib::SignalHandlerId,
```
sdroege, line 219: Why the underscores?

protocol/src/lib.rs
```diff
@@ -26,10 +42,16 @@ pub enum OutgoingMessage {
...
+    ProducerAdded {
+        peer_id: String,
+        display_name: Option<String>,
+    },
     /// Notifies listeners that a producer was removed
     #[serde(rename_all = "camelCase")]
-    ProducerRemoved { peer_id: String },
+    ProducerRemoved {
+        peer_id: String,
+        display_name: Option<String>,
```
sdroege, line 53: Below you use `#[serde(default)]` on all these but not here. Would be good to keep that consistent

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: Because that new structure introduced warnings I didn't notice at the time of merging. This structure is basically only used to keep a reference on its two members, do you have a suggestion for doing that more elegantly btw? :)

sdroege: `#[allow(dead_code)]` with a comment explaining why

protocol/src/lib.rs
sdroege: So? :)

MathieuDuponchelle: So this is an outgoing message :D

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: heh, works too but not the elegance I was hoping for :)

sdroege: I think the important point is to document why things are strange :)

PR #35, opened 2022-03-25
Webrtcsink encoder setup

MathieuDuponchelle:


plugins/src/webrtcsink/imp.rs
```diff
@@ -2551,6 +2578,27 @@ impl ObjectImpl for WebRTCSink {
...
+                 * False if webrtcsink should apply a default configuration
+                 */
+                glib::subclass::Signal::builder(
+                    "encoder-setup",
+                    &[
+                        String::static_type().into(),
+                        String::static_type().into(),
+                        gst::Element::static_type().into(),
+                    ],
+                    bool::static_type().into(),
```
sdroege, line 2599: You probably want to
  - Use an accumulator function that causes the next handler to not be called anymore if `false` (or `true`, your decision) is returned
  - Have a default class handler that does the default setup (i.e. the `configure_encoder()` function)

That way applications can connect signal handlers before or after the class handler, and can decide if they just want to extend the behaviour or don't want the class handler to be called. Just like many such signals work in GTK for example.

plugins/examples/webrtcsink-stats-server.rs
```diff
@@ -133,6 +133,24 @@ async fn run(args: Args) -> Result<(), Error> {
         .by_name("ws")
         .unwrap();
 
+    ws.connect("encoder-setup", false, |values| {
+        let encoder = values[3].get::<gst::Element>().unwrap();
+
+        info!("Encoder: {}", encoder.factory().unwrap().name());
+
+        let configured = match encoder.factory().unwrap().name().as_str() {
```
sdroege, line 141: You maybe want to get the factory (name) only once here, and also not panic if the element has no factory

plugins/examples/webrtcsink-stats-server.rs
```diff
@@ -133,6 +133,24 @@ async fn run(args: Args) -> Result<(), Error> {
         .by_name("ws")
         .unwrap();
 
+    ws.connect("encoder-setup", false, |values| {
```
sdroege, line 136: This `false` here makes it connect *before* the class handler btw, a `true` would be after.

PR #36, opened 2022-03-25
webrtcsink: turn NavigationEventHandler into a tuple wrapper

MathieuDuponchelle:
More elegant way of not bumping into the unused fields warning
than prefixing with _


MathieuDuponchelle:
cc @sdroege :)

sdroege:
Well that works but is a bit strange :) You should've added a comment why it is like this

PR #38, opened 2022-03-25
webrtcsink: address Sebastian's latest comments

MathieuDuponchelle:
+ refactor encoder-setup signal emission
+ document unused fields in navigation event handler


plugins/src/webrtcsink/imp.rs
```diff
@@ -2595,6 +2586,23 @@ impl ObjectImpl for WebRTCSink {
                     ],
                     bool::static_type().into(),
                 )
+                .accumulator(|_hint, _ret, value| !value.get::<bool>().unwrap())
```
sdroege, line 2589: You probably also want to set `_ret` here. See the implementation of `g_signal_accumulator_true_handled()`, which is basically this here.

PR #39, opened 2022-03-30
Add property to specify certificate authority

compiaffe:
Under certain circumstances one would like to use a self signed
certificate without the obstacles many browsers implement to
safeguard their users. One approach is to create a certificate
authority. One can now add this to the OS certificate store, or simply
add it to the browser. In the latter case one also needs to specify the
new authority on the webrtcsink. This commit enables this.


compiaffe:
FYI: @Hejsil

MathieuDuponchelle:
fwiw I use https://github.com/FiloSottile/mkcert :)

plugins/src/signaller/imp.rs
```diff
@@ -347,13 +357,22 @@ impl ObjectSubclass for Signaller {
...
+                glib::ParamSpecString::new(
+                    "address",
+                    "Address",
+                    "Address of the signalling server",
+                    Some("ws://127.0.0.1:8443"),
+                    glib::ParamFlags::READWRITE,
+                ),
+                glib::ParamSpecString::new(
+                    "cafile",
+                    "cafile",
```
MathieuDuponchelle, line 370: CA File

plugins/src/signaller/imp.rs
```diff
@@ -347,13 +357,22 @@ impl ObjectSubclass for Signaller {
...
+                    "cafile",
```
MathieuDuponchelle, line 371: Path to a Certificate file to add to the set of roots the TLS connector will trust

MathieuDuponchelle:
lgtm otherwise, thanks :)

compiaffe:
@MathieuDuponchelle done :)  mkcert looks good, I bash this ATM.

PR #41, opened 2022-04-12
Make our "loss based control" algorithm closer to what is defined in …

thiblahute:
As specified in GCC we should run the packet loss estimation algorithm
when we receive the feedback.

This also starts making use of the feedback packet loss computation
instead of twcc one as it is more accurate since it should take into
account reordering. We aggregate the packet loss from the difference
sources weighting it with the number packets that have been tried to be
sent over the network since the last feedback.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -120,14 +120,21 @@ struct VideoEncoder {
 }
 
 struct CongestionController {
-    /// Overall bitrate target for all video streams.
+    /// Note: The target bitrate applied is the min of
+    /// target_bitrate_on_delay and target_bitrate_on_loss
+    ///
+    /// Bitrate target based on delay factor for all video streams.
     /// Hasn't been tested with multiple video streams, but
     /// current design is simply to divide bitrate equally.
-    bitrate_ema: Option<f64>,
+    target_bitrate_on_delay: i32,
+
+    /// Bitrate target based on loss for all video streams.
+    target_bitrate_on_loss: i32,
+
     /// Exponential moving average, updated when bitrate is
     /// decreased, discarded when increased again past last
     /// congestion window. Smoothing factor hardcoded.
-    target_bitrate: i32,
+    bitrate_ema: Option<f64>,
     /// Exponentially weighted moving variance, recursively
     /// updated along with bitrate_ema. sqrt'd to obtain standard
     /// deviation, used to determine whether to increase bitrate
@@ -156,9 +163,21 @@ enum CongestionControlOp {
     /// Don't update target bitrate
     Hold,
     /// Decrease target bitrate
-    Decrease(f64),
+    Decrease {
+        factor: f64,
+        #[allow(dead_code)]
+        reason: String, // for Debug
+    },
     /// Increase target bitrate, either additively or multiplicatively
     Increase(IncreaseType),
+}
+
+#[derive(Debug, Clone, Copy)]
+enum ControllerType {
+    // Running the "delay-based controller"
+    Delay,
+    // Running the "loss based controller"
+    Loss,
 }
 
 struct Consumer {
@@ -173,6 +192,8 @@ struct Consumer {
     stats: gst::Structure,
 
     max_bitrate: u32,
+
+    stats_sigid: Option<glib::SignalHandlerId>,
 }
 
 #[derive(PartialEq)]
@@ -209,6 +230,7 @@ fn create_navigation_event<N: IsA<gst_video::Navigation>>(sink: &N, msg: &str) {
         gst_error!(CAT, "Invalid navigation event: {:?}", msg);
     }
 }
+
 /// Simple utility for tearing down a pipeline cleanly
 struct PipelineWrapper(gst::Pipeline);
 
@@ -454,20 +476,6 @@ fn setup_encoding(
     Ok((enc, conv_filter, pay))
 }
 
-fn lookup_remote_inbound_rtp_stats(stats: &gst::StructureRef) -> Option<gst::Structure> {
-    for (_, field_value) in stats {
-        if let Ok(s) = field_value.get::<gst::Structure>() {
-            if let Ok(type_) = s.get::<gst_webrtc::WebRTCStatsType>("type") {
-                if type_ == gst_webrtc::WebRTCStatsType::RemoteInboundRtp {
-                    return Some(s);
-                }
-            }
-        }
-    }
-
-    None
-}
-
 fn lookup_transport_stats(stats: &gst::StructureRef) -> Option<gst::Structure> {
     for (_, field_value) in stats {
         if let Ok(s) = field_value.get::<gst::Structure>() {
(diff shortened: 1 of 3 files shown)


plugins/src/webrtcsink/imp.rs
```diff
@@ -1642,24 +1737,50 @@ impl WebRTCSink {
...
+        let rtpbin = webrtcbin
+            .dynamic_cast_ref::<gst::ChildProxy>()
+            .unwrap()
+            .child_by_name("rtpbin")
+            .unwrap();
+
+        if consumer.congestion_controller.is_some() {
+            let peer_id_str = peer_id.to_string();
+            consumer.stats_sigid = Some(rtpbin.connect_closure("on-new-ssrc", true,
+            glib::closure!(@weak-allow-none element, @weak-allow-none webrtcbin
```
sdroege, line 1764: Yay, `glib::closure!` !

plugins/src/webrtcsink/imp.rs
```diff
@@ -1642,24 +1737,50 @@ impl WebRTCSink {
...
+                            => move |rtpbin: gst::Object, session_id: u32, _src: u32| {
+                    let session = rtpbin.emit_by_name::<gst::Element>("get-session", &[&session_id]).property::<glib::Object>("internal-session");
+
+                    let element = element.expect("on-new-ssrc emited when webrtcsink has been disposed?");
+                    let webrtcbin = webrtcbin.unwrap();
+                    let mut state = element.imp().state.lock().unwrap();
+                    if let Some(mut consumer) = state.consumers.get_mut(&peer_id_str) {
+                        consumer.stats_sigid = Some(session.connect("on-feedback-rtcp",
+                            false,
+                            glib::clone!(@strong peer_id_str, @weak webrtcbin, @weak element => @default-panic, move |_| {
```
sdroege, line 1774: But why not also here?

plugins/src/webrtcsink/imp.rs
```diff
@@ -120,14 +120,21 @@ struct VideoEncoder {
 }
 
 struct CongestionController {
-    /// Overall bitrate target for all video streams.
+    /// Note: The target bitrate applied is the min of
+    /// target_bitrate_on_delay and target_bitrate_on_loss
```
MathieuDuponchelle, line 124: Why? :)

thiblahute:
```
   The loss-based estimate As_hat is compared with the delay-based
   estimate A_hat.  The actual sending rate is set as the minimum
   between As_hat and A_hat.
```

plugins/src/webrtcsink/imp.rs
thiblahute: Because I didn't care at all about the arguments in that case.

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: Heh, cool :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -454,18 +475,46 @@ fn setup_encoding(
...
+            s.get::<f64>("fraction-lost")
+                .map(|f_lost| {
+                    sum_fraction_lost += f_lost;
+                    n_fraction_losts += 1.;
+                })
+                .ok();
+        })
+        .collect::<()>();
+
+    let loss_percentage = (sum_fraction_lost / n_fraction_losts) * 100.;
```
MathieuDuponchelle, line 514: This is related to what we discussed, this is not correct, you can't just make an average of loss percentages to calculate the overall loss percentage :)

thiblahute: I removed the weight, but I see you were just ignoring it all for the RTT and were using only the first value, so this is more correct in any case.

MathieuDuponchelle: Not sure it matters whether things were incorrect before, what we want here is to operate on the absolute numbers, and track those from one report to the next ;)

thiblahute: The stat is just not present here, using the percentage is correct here.

thiblahute:
This now depends on https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2266 - I am not sure what we want to do.

plugins/src/webrtcsink/imp.rs
```diff
@@ -787,29 +779,132 @@ impl CongestionController {
...
+        }
+
+        inbound_rtp_stats
+    }
+
+    fn lookup_rtt(&self, stats: &gst::StructureRef) -> f64 {
+        let inbound_rtp_stats = self.get_remote_inbound_stats(stats);
+        let mut rtt = 0.;
+        let mut n_rtts = 0u64;
+        for inbound_stat in inbound_rtp_stats.iter() {
```
sdroege, line 819:
```suggestion
        for inbound_stat in &inbound_rtp_stats {
```

I see my suggestion got rid of the christmas tree below :)

thiblahute: Yeah, thanks ;)

plugins/src/webrtcsink/imp.rs
```diff
@@ -787,81 +759,187 @@ impl CongestionController {
...
         stats: &gst::StructureRef,
         encoders: &mut Vec<VideoEncoder>,
     ) {
-        let n_encoders = encoders.len() as i32;
+        let loss_percentage = stats.get::<f64>("packet-loss-pct").unwrap();
 
-        let rtt = lookup_remote_inbound_rtp_stats(stats)
-            .and_then(|s| s.get::<f64>("round-trip-time").ok())
-            .unwrap_or(0.);
+        gst_error!(CAT, "Loss percentage: {}", loss_percentage);
```
MathieuDuponchelle, line 825: I know we're happy to have found the value, but maybe we don't need to proclaim it that loudly :)

MathieuDuponchelle, line 825: I know we're happy to have found the value, but maybe we don't need to proclaim it that loudly :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -787,81 +759,187 @@ impl CongestionController {
...
+                );
 
-            let target_bitrate = self.target_bitrate / n_encoders;
-
-            let fec_ratio = {
-                if target_bitrate <= 2000000 || self.max_bitrate <= 2000000 {
-                    0f64
+                // Smoothing factor
+                let alpha = 0.75;
+                if let Some(ema) = self.bitrate_ema {
```
MathieuDuponchelle, line 900: I'm a bit confused, I thought part of the benefits of this patch was that the loss-based controller would *not* update the detected congestion window (if any)? Shouldn't we only update `bitrate_ema` and `bitrate_emvar` when the controller_type is Delay ?

MathieuDuponchelle, line 900: I'm a bit confused, I thought part of the benefits of this patch was that the loss-based controller would *not* update the detected congestion window (if any)? Shouldn't we only update `bitrate_ema` and `bitrate_emvar` when the controller_type is Delay ?

MathieuDuponchelle:
Nearly OK with the patch otherwise, looks like a very significant improvement :)

plugins/src/webrtcsink/imp.rs
thiblahute: Indeed, my mistake.

thiblahute: I wonder why your comments are duplicated :-)

MathieuDuponchelle: lol yeah, not sure, I didn't hit Enter twice is all I can say :)

MathieuDuponchelle:
Thanks!

PR #42, opened 2022-04-12
sink: Force data channel priority to high

thiblahute:
Low latency is very important for user interaction

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -1226,9 +1226,16 @@ impl InputStream {
 
 impl NavigationEventHandler {
     pub fn new(element: &super::WebRTCSink, webrtcbin: &gst::Element) -> Self {
+        gst_info!(CAT, "Creating navigation data channel");
         let channel = webrtcbin.emit_by_name::<WebRTCDataChannel>(
             "create-data-channel",
-            &[&"input", &None::<gst::Structure>],
+            &[
+                &"input",
+                &gst::Structure::new(
+                    "config",
+                    &[("priority", &gst_webrtc::WebRTCPriorityType::High)],
+                ),
+            ],
         );
 
         let weak_element = element.downgrade();


plugins/src/webrtcsink/imp.rs
```diff
@@ -1226,18 +1219,51 @@ impl InputStream {
...
+
+                        let root: serde_json::Value = serde_json::from_str(msg)
+                            .expect("We already parsed the json as GstEvent, that shouldn't fail");
+                        let value: Option<u64> =
+                            root.get("timestamp")
+                                .and_then(|value| value.as_u64());
+
+                        match value {
+                            Some(timestamp) => {
+                                channel.send_string(Some(&json!({
```
sdroege, line 1253: The web parts here are not updated to actually set this, and this should probably get some documentation somewhere too

thiblahute: I think it is good to let the client measure some stats, I will add support for that in the client .js lib.

MathieuDuponchelle: As I said, I don't think it is generic enough to warrant inclusion :)

thiblahute: Not sure what kind of genericity you are talking about here tbh.

MathieuDuponchelle: Well, you're just making up a new event type from scratch here, and expecting fields in an existing event as well, that's the kind of genericity I'm talking about ;) For the record, I'm happy with the current navigation interface because the event that gets transported maps to an actual GStreamer event, without modification.

thiblahute: It is a generic feedback mechanism for any gst events sent, obviously it doesn't map other way of using Gstreamer interface since we are implementing a specific use case here. The [`timestamp`](https://gstreamer.freedesktop.org/documentation/gstreamer/gstevent.html?gi-language=c#members) field exists in `GstEvent` itself, I am not making this up.

MathieuDuponchelle: heh, didn't know about that timestamp field indeed, but as far as I can tell it's hardly used anywhere and not really well specified.

thiblahute: Indeed, but " the timestamp of the event" matches perfectly here :P

Tbh I can cary on that patch outside the repo, but I do think it is generically useful.

MathieuDuponchelle: I'd rather you did yes :)

thiblahute: hrm, done...

MathieuDuponchelle:
thanks :)

PR #43, opened 2022-04-14
Add touch events and modifier states

vivienne-w:
This adds support for the new touch screen events and modifier states introduced in gst 1.22. Tested in chromium.

### DIFF
--- a/plugins/Cargo.toml
+++ b/plugins/Cargo.toml
@@ -9,12 +9,12 @@ repository = "https://github.com/centricular/webrtcsink/"
 build = "build.rs"
 
 [dependencies]
-gst = { package = "gstreamer", version = "0.18", features = ["v1_20", "ser_de"] }
-gst-app = { package = "gstreamer-app", version = "0.18", features = ["v1_20"] }
-gst-video = { package = "gstreamer-video", version = "0.18", features = ["v1_20", "ser_de"] }
-gst-webrtc = { package = "gstreamer-webrtc", version = "0.18", features = ["v1_20"] }
-gst-sdp = { package = "gstreamer-sdp", version = "0.18", features = ["v1_20"] }
-gst-rtp = { package = "gstreamer-rtp", version = "0.18", features = ["v1_20"] }
+gst = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package = "gstreamer", features = ["v1_20", "ser_de"] }
+gst-app = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package = "gstreamer-app", features = ["v1_20"] }
+gst-video = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package = "gstreamer-video", features = ["v1_20", "ser_de"] }
+gst-webrtc = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package = "gstreamer-webrtc", features = ["v1_20"] }
+gst-sdp = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package = "gstreamer-sdp", features = ["v1_20"] }
+gst-rtp = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package = "gstreamer-rtp", features = ["v1_20"] }
 once_cell = "1.0"
 smallvec = "1"
 anyhow = "1"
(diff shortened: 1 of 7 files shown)


plugins/Cargo.toml
```diff
@@ -9,12 +9,12 @@ repository = "https://github.com/centricular/webrtcsink/"
 build = "build.rs"
 
 [dependencies]
-gst = { package = "gstreamer", version = "0.18", features = ["v1_20", "ser_de"] }
-gst-app = { package = "gstreamer-app", version = "0.18", features = ["v1_20"] }
-gst-video = { package = "gstreamer-video", version = "0.18", features = ["v1_20", "ser_de"] }
-gst-webrtc = { package = "gstreamer-webrtc", version = "0.18", features = ["v1_20"] }
-gst-sdp = { package = "gstreamer-sdp", version = "0.18", features = ["v1_20"] }
-gst-rtp = { package = "gstreamer-rtp", version = "0.18", features = ["v1_20"] }
+gst = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", rev="61a3f5296c3093e9811d883256ad38bd5483b6db", package = "gstreamer", features = ["v1_22", "ser_de"] }
```
thiblahute, line 12: I do not think we should always depend on 1.22 but instead have it as an option of this crate.

www/input.js
```diff
@@ -404,10 +466,10 @@ class Input {
         // Using guacamole keyboard because it has the keysym translations.
         this.keyboard = new Keyboard(window);
         this.keyboard.onkeydown = (keysym) => {
-            this.send(JSON.stringify( {"event": "KeyPress", "key": keysym}));
+            this.send(JSON.stringify( {"event": "KeyPress", "key": keysym, "modifier_state": ""}));
```
thiblahute, line 469: Add a FIXME here at least, we should be able to have modifiers info.

www/input.js
```diff
@@ -171,42 +183,87 @@ class Input {
...
+        event.preventDefault();
     }
 
     /**
      * Handles mouse wheel events and sends them to WebRTC app.
      * @param {MouseEvent} event
      */
     _wheel(event) {
+        let masks = [];
+        if (event.altKey) mods += "alt-mask";
```
thiblahute, line 255: Factor out a function for that?

plugins/Cargo.toml
```diff
@@ -9,12 +9,12 @@ repository = "https://github.com/centricular/webrtcsink/"
 build = "build.rs"
 
 [dependencies]
-gst = { package = "gstreamer", version = "0.18", features = ["v1_20", "ser_de"] }
-gst-app = { package = "gstreamer-app", version = "0.18", features = ["v1_20"] }
-gst-video = { package = "gstreamer-video", version = "0.18", features = ["v1_20", "ser_de"] }
-gst-webrtc = { package = "gstreamer-webrtc", version = "0.18", features = ["v1_20"] }
-gst-sdp = { package = "gstreamer-sdp", version = "0.18", features = ["v1_20"] }
-gst-rtp = { package = "gstreamer-rtp", version = "0.18", features = ["v1_20"] }
+gst = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", rev="61a3f5296c3093e9811d883256ad38bd5483b6db", package = "gstreamer", features = ["v1_20", "ser_de"] }
```
MathieuDuponchelle, line 12: @sdroege does that seem OK to you? :)

thiblahute: We depend on `gstreamer-rs` main, but decided to pin a specific commit here.

sdroege: Well if you want to keep this up to date :shrug:

MathieuDuponchelle: ok thanks, I think given webrtcsink should eventually live in `gst-plugins-rs`, we probably should do what's done there and just depend on main.

sdroege: You probably want to update after https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/1019 is merged

MathieuDuponchelle: @vivienne-w let's just use main :)

MathieuDuponchelle: (ie you can remove the `rev` altogether)

MathieuDuponchelle:
@thiblahute apart from the build system remarks I have nothing relevant to say, can you take over review? :)

plugins/Cargo.toml
thiblahute: We still want to pin in the `.lock` file I believe.

thiblahute:
> @thiblahute apart from the build system remarks I have nothing relevant to say, can you take over review? :)

lgtm.

MathieuDuponchelle:
thanks!

PR #45, opened 2022-04-22
Consumer pipeline should use_clock

compiaffe:
As per private issue https://github.com/aivero-support/centricular-consulting/issues/26#issuecomment-1106602437
we should `use_clock()` instead of `set_clock()` on the consumer pipeline


compiaffe:
@MathieuDuponchelle @sdroege

compiaffe:
FYI @hejsil

sdroege (approved):

MathieuDuponchelle:
thanks :)

PR #47, opened 2022-04-22
Remove erranous unwrap() on use_clock

compiaffe:
Overlooked a superfluous unwrap() that caused the build to fail. Now it builds.
Also implemented all clippy suggestions.


compiaffe:
@sdroege @MathieuDuponchelle

sdroege (approved):

PR #51, opened 2022-04-27
Add support for VAAPI h.264 and VP8 encoder

### LINKED ISSUE #50: Panics when trying to use VAAPI plugins
After obtaining the vappi plugins (`GstVaapiEncodeVP8`, `GstVaapiEncodeH264` etc.) the server panics upon an incoming connection.

The last log output from gstreamer before it dies is: `webrtcsink plugins/src/webrtcsink/imp.rs:2598:webrtcsink::webrtcsink::imp:<rswebrtcsink0> applying default configuration on encoder Element { inner: ObjectRef { inner: 0x7f67d40f05a0, type: GstVaapiEncodeVP8 } }`

```
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: webrtcsink::webrtcsink::imp::VideoEncoder::bitrate
             at ./plugins/src/webrtcsink/imp.rs:514:18
   4: webrtcsink::webrtcsink::imp::Consumer::connect_input_stream
             at ./plugins/src/webrtcsink/imp.rs:1130:57
   5: webrtcsink::webrtcsink::imp::WebRTCSink::on_remote_description_set
             at ./plugins/src/webrtcsink/imp.rs:1819:25
   6: webrtcsink::webrtcsink::imp::WebRTCSink::handle_sdp::{{closure}}
             at ./plugins/src/webrtcsink/imp.rs:1981:21
   7: gstreamer::promise::Promise::with_change_func::trampoline
             at /home/septatrix/.cargo/registry/src/github.com-1ecc6299db9ec823/gstreamer-0.18.6/src/promise.rs:67:13
   8: <unknown>
   9: <unknown>
  10: g_main_context_dispatch
  11: <unknown>
  12: g_main_loop_run
  13: <unknown>
  14: <unknown>
  15: start_thread
  16: __GI___clone3
```

[Log output](https://github.com/centricular/webrtcsink/files/8556956/webrtc.log) from `GST_PLUGIN_PATH=$PWD/target/debug GST_DEBUG="webrtc*:9,vaapi*:9" RUST_BACKTRACE=1 gst-launch-1.0 videotestsrc ! webrtcsink &>webrtc.log`

septatrix:
This fixes #50.
I am not sure what the effect of the other options in `configure_encoder` is so other improvements for tweaking latency etc might be possible.
Additionally I added a reason inside the `unreachable!` macro though I guess semantically a panic or whatever mechanism `gstreamer-rs` exposes to signal an error would be better.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -330,6 +330,9 @@ fn configure_encoder(enc: &gst::Element) {
                 enc.set_property_from_str("rc-mode", "cbr-ld-hq");
                 enc.set_property("zerolatency", true);
             }
+            "vaapih264enc" => {
+                enc.set_property("bitrate", 2048u32);
+            }
             _ => (),
         }
     }
@@ -510,8 +513,8 @@ impl VideoEncoder {
     fn bitrate(&self) -> i32 {
         match self.factory_name.as_str() {
             "vp8enc" | "vp9enc" => self.element.property::<i32>("target-bitrate"),
-            "x264enc" | "nvh264enc" => (self.element.property::<u32>("bitrate") * 1000) as i32,
-            _ => unreachable!(),
+            "x264enc" | "nvh264enc" | "vaapih264enc" => (self.element.property::<u32>("bitrate") * 1000) as i32,
+            factory => unreachable!("Factory {} is currently not supported", factory),
         }
     }
 
@@ -532,10 +535,10 @@ impl VideoEncoder {
     fn set_bitrate(&mut self, element: &super::WebRTCSink, bitrate: i32) {
         match self.factory_name.as_str() {
             "vp8enc" | "vp9enc" => self.element.set_property("target-bitrate", bitrate),
-            "x264enc" | "nvh264enc" => self
+            "x264enc" | "nvh264enc" | "vaapih264enc" => self
                 .element
                 .set_property("bitrate", (bitrate / 1000) as u32),
-            _ => unreachable!(),
+            factory => unreachable!("Factory {} is currently not supported", factory),
         }
 
         let current_caps = self.filter.property::<gst::Caps>("caps");
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -333,6 +333,7 @@ fn configure_encoder(enc: &gst::Element) {
             "vaapih264enc" | "vaapivp8enc" => {
                 enc.set_property("bitrate", 2048u32);
                 enc.set_property("keyframe-period", 2560u32);
+                enc.set_property_from_str("rate-control", "cbr");
             }
             _ => (),
         }
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -330,8 +330,9 @@ fn configure_encoder(enc: &gst::Element) {
                 enc.set_property_from_str("rc-mode", "cbr-ld-hq");
                 enc.set_property("zerolatency", true);
             }
-            "vaapih264enc" => {
+            "vaapih264enc" | "vaapivp8enc" => {
                 enc.set_property("bitrate", 2048u32);
+                enc.set_property("keyframe-period", 2560u32);
             }
             _ => (),
         }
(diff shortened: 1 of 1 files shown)


sdroege (approved):

septatrix:
> Additionally I added a reason inside the `unreachable!` macro though I guess semantically a panic or whatever mechanism `gstreamer-rs` exposes to signal an error would be better.

Regarding the last point: There are `element_error`, `error_msg`, `gst_error`, `loggable_error` and `panic_to_error` (and probably some more) though they are not well documented. Might one of these be more appropriate than the `unreachable!`? Or should this just be left as is until encoder whitelists are implemented such that this would be a real "unreachable" match?

sdroege:
It's more `unimplemented!()` / `todo!()` than `unreachable!()` in any case :)

If you want to handle this gracefully you probably want to post an error message from the sink.

MathieuDuponchelle:
`unimplemented()` would make more sense for the time being yes.

plugins/src/webrtcsink/imp.rs
```diff
@@ -510,8 +513,8 @@ impl VideoEncoder {
     fn bitrate(&self) -> i32 {
         match self.factory_name.as_str() {
             "vp8enc" | "vp9enc" => self.element.property::<i32>("target-bitrate"),
```
MathieuDuponchelle, line 515: were you able to test with the vaapi VP8 encoder ooc?

MathieuDuponchelle:
> I am not sure what the effect of the other options in configure_encoder is so other improvements for tweaking latency etc might be possible.

I suppose it's good already if you see things behaving as expected with those basic settings, but yes looking at the doc I think we might want to tune the `quality-level` property in the future, depending on what latency we observe with the default value.

One thing we can do right now is set `keyframe-period` to the same (very high) value as with the other encoders (`2560`): the exact value doesn't really matter, but that should make bitrate spikes less frequent, which helps with congestion control. I would also probably set `rate-control` to `cbr`.

plugins/src/webrtcsink/imp.rs
septatrix: No I did not test it as hardware accelerated video decoding in Browsers under Linux is always a PITA though I guess I could at least check if it works

septatrix:
> `unimplemented()` would make more sense for the time being yes.

Then I will change that.

> > I am not sure what the effect of the other options in configure_encoder is so other improvements for tweaking latency etc might be possible.
> 
> I suppose it's good already if you see things behaving as expected with those basic settings, but yes looking at the doc I think we might want to tune the `quality-level` property in the future, depending on what latency we observe with the default value.
> 
> One thing we can do right now is set `keyframe-period` to the same (very high) value as with the other encoders (`2560`): the exact value doesn't really matter, but that should make bitrate spikes less frequent, which helps with congestion control. I would also probably set `rate-control` to `cbr`.

I can try to get `comcast` working and check what the behaviour is in extreme cases. Latency seemed good but maybe there are some settings which result in visible improvements, might also check that out.

Do you know any tool which I can use to look out for latency spikes? `iftop` does not really refresh fast enough I would assume.

plugins/src/webrtcsink/imp.rs
```diff
@@ -330,6 +330,10 @@ fn configure_encoder(enc: &gst::Element) {
                 enc.set_property_from_str("rc-mode", "cbr-ld-hq");
                 enc.set_property("zerolatency", true);
             }
+            "vaapih264enc" | "vaapivp8enc" => {
+                enc.set_property("bitrate", 2048u32);
+                enc.set_property("keyframe-period", 2560u32);
+            }
```
septatrix, line 336: I noticed that the default values for bitrate and keyframe distance differ between h264 and VP8 (taking the different unit  (Bit/s, KBit/s into account) even though bitrate for example is set for both encodings the same way in latter functions. However as both encoders have the same properties I grouped them here.

The latency seemed rather low though I only checked the delay between my PC clock and `clockoverlay` as I did not have any better tooling for latency evaluation.
I am not sure if setting constant bitrate (`cbr`) mode yields any benefits as this denies the codec the possibility to use a lower bitrate if feasible (at least that is my understanding—again I do not know too much about gstreamer or encoders ein general)

MathieuDuponchelle:
> Do you know any tool which I can use to look out for latency spikes? `iftop` does not really refresh fast enough I would assume.

Latency is not the thing I'm worried to see spiking, rather actual bitrate coming out of the encoder.

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: So, the idea with the implementation of congestion control is that the algorithm is constantly trying to probe the available bandwidth by raising the bitrate (slowly when it considers that it is within a window where congestion was previously observed, then more rapidly once out of that window). This doesn't work well if for some reason (eg static scene or what have you) the encoder decides it doesn't actually need to use as much bitrate as requested, and that's why using CBR can help by making things more predictable :)

septatrix: Ah that makes sense. I added that setting now.

MathieuDuponchelle:
Thanks!

PR #52, opened 2022-04-28
Remove is_video field from Codec struct

septatrix:
Instead add it as a function and get the value from the encoder type.
This removes the necessity for any manual bookkeeping.

As per discussion in https://github.com/centricular/webrtcsink/issues/50#issuecomment-1111122804.


sdroege (approved):

MathieuDuponchelle:
Thanks!

PR #53, opened 2022-04-28
Remove cd in usage instructions

septatrix:
The `cd` did not serve any purpose which the pythons server does not already have so instead use that to save an instruction.
Instructions could be made even simpler when using background processes though that would hide any logs...

```shell
# Can be copy-pasted into a single shell
cargo run --bin server &>/dev/null &
python3 -m http.server -d www/ &>/dev/null &
gst-launch-1.0 --gst-plugin-path=$PWD v4l2src ! clockoverlay ! webrtcsink stun-server=null
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,7 @@ WEBRTCSINK_SIGNALLING_SERVER_LOG=debug cargo run --bin server
 In the second, run:
 
 ``` shell
-cd www
-python3 -m http.server
+python3 -m http.server -d www/
 ```
 
 In the third, run:


MathieuDuponchelle:
worksforme, note that my plan to simplify these instructions in a satisfying manner would be to instead expose properties to let the element itself instantiate a simple http server, and instantiate the signalling server (that's made possible by the fact that I recently reimplemented it in rust :))

septatrix:
> worksforme, note that my plan to simplify these instructions in a satisfying manner would be to instead expose properties to let the element itself instantiate a simple http server, and instantiate the signalling server (that's made possible by the fact that I recently reimplemented it in rust :))

I gotta say that I actually quite like the way everything is decoupled currently. Also makes it a lot easier to set up some sandboxing and access controls.

MathieuDuponchelle:
> I gotta say that I actually quite like the way everything is decoupled currently. Also makes it a lot easier to set up some sandboxing and access controls.

Yes, it would remain so, the properties would be optional :)

PR #66, opened 2022-06-06
plugin: Implement a bandwidth estimator based on the Google Congestion Control algorithm

thiblahute:
See https://datatracker.ietf.org/doc/html/draft-ietf-rmcat-gcc-02
    
This commit implements the bandwidth estimation as a GStreamer element
that is then used in webrtcbin through the new `request-bandwidth-estimator`
signal.


plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+        let mut list = gst::BufferList::new();
+        let mutlist = list.get_mut().unwrap();
+
+        // Leak the bucket so it can hold at most 30ms of data
+        let maximum_remaining_bits = 30. * self.estimated_bitrate as f64 / 1000.;
+        while (budget > 0 || remaining > maximum_remaining_bits) && !self.buffers.is_empty() {
+            let buf = self.buffers.pop_back().unwrap();
+            let n_bits = buf.size() * 8;
+
+            mutlist.add(buf);
```
ystreet, line 713: Does this rely on basesink/nicesink not handling waiting for each buffer in a buffer list?

plugins/Cargo.toml
```diff
@@ -17,6 +17,7 @@ gst-sdp = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package
 gst-rtp = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package = "gstreamer-rtp", features = ["v1_20"] }
 gst-utils = { git="https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", package = "gstreamer-utils" }
 once_cell = "1.0"
+chrono = "0.4"
```
sdroege, line 20:
```suggestion
chrono = { version = "0.4", default-features = false }
```

plugins/src/gcc/mod.rs
```diff
@@ -0,0 +1,19 @@
+use gst::glib;
+use gst::prelude::*;
+mod imp;
+
+glib::wrapper! {
+    pub struct BandwidthEstimator(ObjectSubclass<imp::BandwidthEstimator>) @extends gst::Element, gst::Object;
+}
+
+unsafe impl Send for BandwidthEstimator {}
+unsafe impl Sync for BandwidthEstimator {}
```
sdroege, line 10: This is not needed anymore with the git version of the glib bindings (can you also remove the others of this?)

plugins/src/webrtcsink/imp.rs
```diff
@@ -2067,13 +1664,23 @@ impl WebRTCSink {
...
-                    let peer_id_clone = peer_id_clone.clone();
                     if let (Some(webrtcbin), Some(element)) =
                         (webrtcbin.upgrade(), element_clone.upgrade())
                     {
-                        element
-                            .imp()
-                            .process_stats(&element, webrtcbin, &peer_id_clone);
+                            let peer_id = peer_id.to_string();
+                            let promise = gst::Promise::with_change_func(
+                                glib::clone!(@strong peer_id, @weak element => move |reply| {
```
sdroege, line 1672:
```suggestion
                                glib::clone!(@weak element => move |reply| {
```

Can just be moved in there or not?

plugins/src/webrtcsink/imp.rs
```diff
@@ -1970,39 +1578,28 @@ impl WebRTCSink {
...
-                    let mut state = element.imp().state.lock().unwrap();
-                    if let Some(mut consumer) = state.consumers.get_mut(&peer_id) {
-                        if let Some(congestion_controller) = consumer.congestion_controller.as_mut() {
-                            congestion_controller.delay_control(&element, stats, &mut consumer.encoders,);
-                        }
-                        consumer.stats = stats.to_owned();
-                    }
+        if let Some(consumer) = state.consumers.get_mut(peer_id) {
+            let fec_ratio = {
+                if bitrate <= 2_000_000 || consumer.cc_info.max_bitrate <= 2_000_000 {
```
sdroege, line 1586: How is that related to 2MBit/s? Maybe worth a comment :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -1280,16 +879,14 @@ impl Consumer {
...
-                enc.transceiver.set_property("fec-percentage", 0u32);
-            } else {
-                /* If congestion control is disabled, we simply use the highest
-                 * known "safe" value for the bitrate. */
-                enc.set_bitrate(element, self.max_bitrate as i32);
-                enc.transceiver.set_property("fec-percentage", 50u32);
+            match self.cc_info.heuristic {
+                WebRTCSinkCongestionControl::Disabled => {
+                    /* If congestion control is disabled, we simply use the highest
+                     * known "safe" value for the bitrate. */
```
sdroege, line 885:
```suggestion
                    // If congestion control is disabled, we simply use the highest
                    // known "safe" value for the bitrate.
```

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+        }
+    }
+}
+
+impl ElementImpl for BandwidthEstimator {
+    fn metadata() -> Option<&'static gst::subclass::ElementMetadata> {
+        static ELEMENT_METADATA: Lazy<gst::subclass::ElementMetadata> = Lazy::new(|| {
+            gst::subclass::ElementMetadata::new(
+                "Google Congestion Control bandwidth estimator",
+                "Network/WebRTC/Rtp/Filter",
```
sdroege, line 1233:
```suggestion
                "Network/WebRTC/RTP/Filter",
```

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+                state.estimated_bitrate.to_value()
+            }
+            _ => unimplemented!(),
+        }
+    }
+}
+
+impl GstObjectImpl for BandwidthEstimator {}
+
+impl Default for State {
```
sdroege, line 1205: This seems like a strange place between all the other trait impls for `BandwidthEstimator`. Maybe move it up to the definition of the `State` struct

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+                    "Minimal bitrate to use (in bit/sec) when computing it through the bandwidth estimation algorithm",
+                    1,
+                    u32::MAX as u32,
+                    DEFAULT_MIN_BITRATE,
+                    glib::ParamFlags::READWRITE | gst::PARAM_FLAG_MUTABLE_READY,
+                ),
+                glib::ParamSpecUInt::new(
+                    "max-bitrate",
+                    "Minimal Bitrate",
+                    "Minimal bitrate to use (in bit/sec) when computing it through the bandwidth estimation algorithm",
```
sdroege, line 1145:
```suggestion
                    "Maximal Bitrate",
                    "Maximal bitrate to use (in bit/sec) when computing it through the bandwidth estimation algorithm",
```

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+
+        obj.add_pad(&self.sinkpad).unwrap();
+        obj.add_pad(&self.srcpad).unwrap();
+    }
+
+    fn properties() -> &'static [glib::ParamSpec] {
+        static PROPERTIES: Lazy<Vec<glib::ParamSpec>> = Lazy::new(|| {
+            vec![
+                /*
+                 *  gcc:bitrate:
```
sdroege, line 1117: There is no property called `bitrate`

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+                *CAT,
+                glib::bool_error!("Unsupported pad mode {mode:?}"),
+            ))
+        }
+    }
+}
+
+#[glib::object_subclass]
+impl ObjectSubclass for BandwidthEstimator {
+    const NAME: &'static str = "RTPGCCBwE";
```
sdroege, line 1022:
```suggestion
    const NAME: &'static str = "GstRTPGCCBwE";
```

Or so maybe so we add some more namespacing

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+                }
+            }
+            let list = {
+                let mut state = lock_state();
+                state.periodic_push = Some(periodic_id);
+                state.create_buffer_list(&bwe)
+            };
+
+            if list.len() > 0 {
+                let _ = bwe.imp().push_list(list);
```
sdroege, line 986: No need for error handling here, not even logging?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+                    }
+                }
+            }
+            let list = {
+                let mut state = lock_state();
+                state.periodic_push = Some(periodic_id);
+                state.create_buffer_list(&bwe)
+            };
+
+            if list.len() > 0 {
```
sdroege, line 985:
```suggestion
            if !list.is_empty() {
```

If you run clippy it might find some other things

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+
+    fn compare_threshold(&mut self) -> (NetworkUsage, Duration) {
+        const MAX_DELTAS: i64 = 30;
+
+        self.num_deltas += 1;
+        if self.num_deltas < 2 {
+            return (NetworkUsage::Normal, self.estimate);
+        }
+
+        // FIXME: WHere is that delta count comming from?
```
sdroege, line 531:
```suggestion
        // FIXME: Where is that delta count coming from?
```

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+                }
+
+                if self.group.inter_arrival_time_pkt(pkt) < *BURST_TIME
+                    && self.group.inter_delay_variation_pkt(pkt) < Duration::zero()
+                {
+                    self.group.add(*pkt);
+                    continue;
+                }
+
+                let group = mem::replace(&mut self.group, PacketGroup::new());
```
sdroege, line 456:
```suggestion
                let group = mem::take(&mut self.group);
```

If you implement `Default` for `PacketGroup`. Slightly nicer IMHO :)

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+
+    /// Returns the effective received bitrate during the last PACKETS_RECEIVED_WINDOW
+    fn effective_bitrate(&self) -> Bitrate {
+        let len = self.last_received_packets.len();
+        if len == 0 {
+            return 0;
+        }
+
+        let duration = self.last_received_packets.get(len - 1).unwrap().arrival
+            - self.last_received_packets.get(0).unwrap().arrival;
```
sdroege, line 380:
```suggestion
        let duration = self.last_received_packets.back().unwrap().arrival
            - self.last_received_packets.front().unwrap().arrival;
```

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+
+        while last_arrival - self.oldest_packet_in_window_ts() > *PACKETS_RECEIVED_WINDOW {
+            self.last_received_packets.remove(0);
+        }
+    }
+
+    /// Returns the effective received bitrate during the last PACKETS_RECEIVED_WINDOW
+    fn effective_bitrate(&self) -> Bitrate {
+        let len = self.last_received_packets.len();
+        if len == 0 {
```
sdroege, line 375:
```suggestion
        if self.last_received_packets.is_empty() {
```

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+            Err(idx) => idx,
+        };
+        self.last_received_packets.insert(idx, packet);
+        self.evict_old_received_packets();
+    }
+
+    fn evict_old_received_packets(&mut self) {
+        let last_arrival = self
+            .last_received_packets
+            .get(self.last_received_packets.len() - 1)
```
sdroege, line 363:
```suggestion
            .back()
```

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+            .unwrap()
+            .arrival;
+
+        while last_arrival - self.oldest_packet_in_window_ts() > *PACKETS_RECEIVED_WINDOW {
+            self.last_received_packets.remove(0);
```
sdroege, line 368:
```suggestion
            self.last_received_packets.pop_front();
```

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+    fn update_last_received_packets(&mut self, packet: Packet) {
+        let arrival = packet.arrival;
+        let idx = match self
+            .last_received_packets
+            .binary_search_by(|packet| packet.arrival.cmp(&arrival))
+        {
+            Ok(idx) => idx,
+            Err(idx) => idx,
+        };
+        self.last_received_packets.insert(idx, packet);
```
sdroege, line 356: Instead of a `VecDeque` a `BTreeMap<usize, _>` might be more efficient?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+}
+
+fn pdur(d: &Duration) -> String {
+    let stdd = time::Duration::from_nanos(d.num_nanoseconds().unwrap().abs() as u64);
+
+    format!("{}{stdd:?}", if d.lt(&Duration::zero()) { "-" } else { "" })
+}
+
+impl PacketGroup {
+    fn new() -> Self {
```
sdroege, line 177: Or just `#[derive(Default)]`?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
...
+    arrival: Duration,
+    size: usize,
+}
+
+fn human_kbits<T: Into<f64>>(bits: T) -> String {
+    format!("{:.2}kb", (bits.into() / 1_000.))
+}
+
+impl Packet {
+    fn from_structure(structure: &gst::Structure) -> Option<Self> {
```
sdroege, line 131:
```suggestion
    fn from_structure(structure: &gst::StructureRef) -> Option<Self> {
```

Maybe avoid creating a full structure copy where it's used?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
+// Implements https://datatracker.ietf.org/doc/html/draft-ietf-rmcat-gcc-02
+use chrono::Duration;
```
sdroege, line 2: `std::time::Duration` is not sufficient?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1270 @@
+// Implements https://datatracker.ietf.org/doc/html/draft-ietf-rmcat-gcc-02
```
sdroege, line 1:
```suggestion
///! Implements https://datatracker.ietf.org/doc/html/draft-ietf-rmcat-gcc-02
```

sdroege (commented):
Just some style comments, can't really comment on the math

plugins/src/gcc/imp.rs
thiblahute: I already ran it and it was clean (maybe I made some changes betwen the time I ran it and the time I proposed, running it again! :-))

plugins/src/gcc/imp.rs
thiblahute: Renamed it :-)

plugins/src/gcc/imp.rs
thiblahute: Neede to implement manually as chrono::Duration doesn't implement it, but yes it is nicer that way :-)

plugins/src/gcc/imp.rs
thiblahute: I need to read more about the Structure vs StructureRef (and friends) difference, it is still not clear in my head it seems :-)

plugins/src/gcc/imp.rs
thiblahute: What is the `!` for?

plugins/src/gcc/imp.rs
thiblahute: `std::time::Duration` doesn't allow negative duration, and we need it everywhere.

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+    Under,
+}
+
+struct Detector {
+    // Grouping fields
+    group: PacketGroup,
+    prev_group: Option<PacketGroup>,
+    measure: Duration, // Delay variation measure
+
+    // FIXME: This is not used but should be according 5.3:
```
MathieuDuponchelle, line 252: Let's try to re-enable this now that you have something that works :)

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
+///! Implements https://datatracker.ietf.org/doc/html/draft-ietf-rmcat-gcc-02
```
MathieuDuponchelle, line 1: This could use a quick description of the element's design (implements pacing, performs estimation upon receiving downstream events, typically placed in front of `rtpsession`, ..)

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+            let bwe = weak_controller
+                .upgrade()
+                .expect("bwe destroyed while its srcpad task is still running?");
+
+            let lock_state = || bwe.imp().state.lock().unwrap();
+
+            let periodic_id = match lock_state().periodic_push.take() {
+                Some(id) => id,
+                _ => {
+                    gst::info!(CAT, "Peridic pushes are stop, pausing task");
```
MathieuDuponchelle, line 989: weird English here

plugins/src/gcc/imp.rs
sdroege: `Structure` is an owned structure (you're going to free it on `Drop`), `&StructureRef` is a borrowed structure. It's like the difference between `String` and `&str`, `PathBuf` and `&Path`, etc.

For example, if you have a `glib::Value` and get a `Structure` out of it then it's your own copy of that, if you get a `&StructureRef` out of it then you're only borrowing the one owned by the value.

plugins/src/gcc/imp.rs
sdroege: With `!` it applies to the surrounding item (in this case the whole module), without to the following item (e.g. the next struct definition). It's the same thing with attributes, e.g.

```rust
// Applies to everything in this module
#![allow(dead_code)]
struct Foo {
    ...
}

// Applies only to the struct
#[allow(clippy::blabla)]
struct Bar {
    ...
}
```

plugins/src/gcc/imp.rs
sdroege:
```suggestion
                    gst::info!(CAT, "Periodic pushes are stop, pausing task");
```

sdroege: Also a typo

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+            last_control_op: BandwidthEstimationOp::Increase("Initial increase".into()),
+            flow_return: Err(gst::FlowError::Flushing),
+            periodic_push: None,
+            last_push: time::Instant::now(),
+        }
+    }
+}
+
+impl State {
+    fn create_buffer_list(&mut self, bwe: &super::BandwidthEstimator) -> gst::BufferList {
```
MathieuDuponchelle, line 712: Add a comment here pointing out that this is implementing `4. sending engine`, using a "leaky bucket" implementation?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+
+        // Leak the bucket so it can hold at most 30ms of data
+        let maximum_remaining_bits = 30. * self.estimated_bitrate as f64 / 1000.;
+        while (budget > 0 || remaining > maximum_remaining_bits) && !self.buffers.is_empty() {
+            let buf = self.buffers.pop_back().unwrap();
+            let n_bits = buf.size() * 8;
+
+            mutlist.add(buf);
+            budget = budget.saturating_sub(n_bits as i64);
+            remaining -= n_bits as f64;
```
MathieuDuponchelle, line 736: We also want to perform stuffing here right?

plugins/src/gcc/imp.rs
thiblahute: Makes perfect sense, thanks :-)

plugins/src/gcc/imp.rs
MathieuDuponchelle: Using a `VecDeque` here does seem not very adapted (given the requirement for "random" insertion, and the systematic binary search), I don't understand your comment about the seqnum @thiblahute ?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+
+                let group = mem::take(&mut self.group);
+                gst::trace!(
+                    CAT,
+                    "Packet group done: {:?}",
+                    gst::ClockTime::from_nseconds(group.departure.num_nanoseconds().unwrap() as u64)
+                );
+                if let Some(prev_group) = mem::replace(&mut self.prev_group, Some(group.clone())) {
+                    self.last_inter_departure_times
+                        .push(group.inter_departure_time(&prev_group));
```
MathieuDuponchelle, line 462: And this is essentially the result of 5.1 ?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+                    if self.last_inter_departure_times.len() > LAST_INTER_DEPARTURE_SIZE {
+                        self.last_inter_departure_times.remove(0);
+                    }
+                    self.kalman_estimate(&prev_group, &group);
```
MathieuDuponchelle, line 466: 5.3 ?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+                    self.overuse_filter();
```
MathieuDuponchelle, line 467: 5.4 ?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+        // use f_max here.
+        let alpha = ONE_MINUS_CHI.powf(30.0 / (1000. * 5. * 1_000_000.));
+        let root = self.measurement_uncertainty.sqrt();
+        let root3 = 3. * root;
+
+        if zms > root3 {
+            self.measurement_uncertainty =
+                (alpha * self.measurement_uncertainty + (1. - alpha) * root3 * root3).max(1.);
+        }
+        self.measurement_uncertainty =
```
MathieuDuponchelle, line 512: So, if I read the RFC right:

> If z(i) > 3*sqrt(var_v_hat) the filter is updated with 3*sqrt(var_v_hat) rather than z(i).

Shouldn't this be in an else clause?

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+
+    fn compare_threshold(&mut self) -> (NetworkUsage, Duration) {
+        const MAX_DELTAS: i64 = 30;
+
+        self.num_deltas += 1;
+        if self.num_deltas < 2 {
+            return (NetworkUsage::Normal, self.estimate);
+        }
+
+        // FIXME: Where is that delta count coming from?
```
MathieuDuponchelle, line 529: Good question :)

plugins/src/gcc/imp.rs
thiblahute: The synchronization should happen upstream indeed, and the sink should not sync anything.

In `webrtcsink` this is how we do it, but indeed, if the stream is not synced on the clock before webrtcbin it is gonna work because of the suboptimal synchronization  behavior of  buffer lists in BaseSink.

sdroege: Maybe this should come together with a change to split bufferlists somewhere in webrtcbin or webrtcsink

thiblahute: I am not sure to understand your point.

plugins/src/gcc/imp.rs
sdroege: You work with "extended seqnums" for that reason usually (roll-over counter in the upper, extended bits)

plugins/src/gcc/imp.rs
sdroege: If it's a problem that buffer lists are synchronized as one item in sinks instead of pacing each buffer individually then you probably want to make sure that buffer lists are split into individual buffers before the synchronization happens (in the `clocksync` elements here, the sinks instead `webrtcbin` don't sync!).

thiblahute: Well ideally upstream synchronizes before the buffer are put into a list and then we do our pacing without any more synchronization, this is what we do in `webrtcsink`, adding support to sync on individual buffers in `clocksync` could be a good idea too but unrelated to me.

Would you want to add a `clocksync` into `webrtcbin` automatically when using a congestion controller? I do not think it is a good idea. I guess the documentation of `rtpgccbwe` should mention those details though.

sdroege: `webrtcbin` already uses `clocksync` internally to synchronize the separate streams before bundling, but that would still work on full buffer lists at this point.

thiblahute: Ah right, well, not sure it is a problem, especially if the stream has already been synced upstream. Adding proper support to bufferlist sync in clocksync/basesink is anyway an outstanding task fmopv.

plugins/src/gcc/imp.rs
thiblahute: Would need to bring that info into the TWCC stats event/structure I guess. could also probably find a way to handle the rollover counter myself.

plugins/src/gcc/imp.rs
thiblahute: Removed it in the end and added a comment explainning.

plugins/src/gcc/imp.rs
thiblahute: Yes

thiblahute: Yes

plugins/src/gcc/imp.rs
thiblahute: Added a comment explaining we are just being sheep :P

plugins/src/gcc/imp.rs
thiblahute: Implemented.

thiblahute:
This should be ready now @MathieuDuponchelle and @sdroege

plugins/src/gcc/imp.rs
MathieuDuponchelle: So, how was this addressed? :)

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: Not that this is just moved existing code, thibault isn't the guilty party here :)

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1333 @@
...
+
+            let rtt_ms = self.detector.rtt().num_milliseconds() as f64;
+            let response_time_ms = 100. + rtt_ms;
+            let alpha = 0.5 * f64::min(time_since_last_update_ms / response_time_ms, 1.0);
+            let threshold_on_effective_bitrate = 1.5 * effective_bitrate as f64;
+            let increase = f64::max(
+                1000.0f64,
+                f64::min(
+                    alpha * avg_packet_size_bits,
+                    // FIXME: Until we implement "stuffing" which should ensure that the effective bitrate is
```
MathieuDuponchelle, line 822: Where are we on the stuffing front? :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -187,19 +191,19 @@ enum ControllerType {
...
-
 struct Consumer {
     pipeline: gst::Pipeline,
     webrtcbin: gst::Element,
+    rtprtxsend: Option<gst::Element>,
     webrtc_pads: HashMap<u32, WebRTCPad>,
     peer_id: String,
     encoders: Vec<VideoEncoder>,
-    /// None if congestion control was disabled
+    // Our Homegrown controller
```
MathieuDuponchelle, line 201: This is all pretty awkward, I think we should move the implementation of the homegrown CC to a separate file, and name things more explicitly in here, the current state is confusing wrt the gcc controller that exists as an element.

plugins/src/webrtcsink/imp.rs
```diff
@@ -1729,10 +1751,77 @@ impl WebRTCSink {
...
+            WebRTCSinkCongestionControl::GoogleCongestionControl => {
+                let cc_info = settings.cc_info;
+                webrtcbin.connect_closure(
+                    "request-aux-sender",
+                    false,
+                    glib::closure!(@watch element, @strong peer_id
+                            => move |_webrtcbin: gst::Element, _transport: gst::Object| {
+
+                        let cc = match cc_info.heuristic {
+                            WebRTCSinkCongestionControl::GoogleCongestionControl => {
```
MathieuDuponchelle, line 1764: Why do you need to match here? The whole closure is only connected if `cc_info` is gcc isn't it?

plugins/src/webrtcsink/imp.rs
```diff
@@ -1729,10 +1751,77 @@ impl WebRTCSink {
...
+                                match gst::ElementFactory::make("rtpgccbwe", None) {
+                                    Err(err) => {
+                                        glib::g_warning!("webrtcsink",
+                                            "Activated the `rtpgccbwe` element is not available \
```
MathieuDuponchelle, line 1768: hm?

plugins/src/gcc/imp.rs
thiblahute: By ignoring it for now.

plugins/src/gcc/imp.rs
thiblahute: [It is implemented in ](https://github.com/centricular/webrtcsink/pull/66/commits/8836480667368204dd6258a279981363b44054b2)

plugins/src/webrtcsink/imp.rs
thiblahute: Well, moving it away is a bit of work which is not really needed or can be done later fmpov

plugins/src/gcc/imp.rs
thiblahute: It is sent as burst as we don't timestamp the bufferlist that we send, so there in no synchronization happenning after gccwbe.

What could be enhanced is synchronizatin upstream in the `clocksync` that is inside `webrtcbin` or the one before in `webrtcsink` itself.

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1274 @@
...
+        let mutlist = list.get_mut().unwrap();
+
+        // Leak the bucket so it can hold at most 30ms of data
+        let maximum_remaining_bits = 30. * self.estimated_bitrate as f64 / 1000.;
+        while (budget > 0 || remaining > maximum_remaining_bits) && !self.buffers.is_empty() {
+            let buf = self.buffers.pop_back().unwrap();
+            let n_bits = buf.size() * 8;
+
+            mutlist.add(buf);
+            budget = budget.saturating_sub(n_bits as i64);
```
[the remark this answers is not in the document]
thiblahute: We could, complexify a bit the code for afaic ~nothing fmopv

plugins/src/webrtcsink/imp.rs
thiblahute: For futureproofness I guess, I could remove it.

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: Well I disagree :)

plugins/src/webrtcsink/imp.rs
MathieuDuponchelle: please do yes :)

plugins/src/gcc/imp.rs
MathieuDuponchelle: What do you mean "we do not timestamp the bufferlist that we send"? Bufferlists do not have a timestamp, sinks will synchronize (at the moment) by looking at the timestamp of the first buffer only, but there is no guarantee that this will remain true in the future.

Besides, webrtcbin could add elements that split up buffer lists downstream of the estimator for some reason (or might already do so, I haven't checked), so keeping the timestamps unchanged isn't very reassuring here :)

plugins/src/gcc/imp.rs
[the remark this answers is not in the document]
MathieuDuponchelle: Accuracy is not nothing ..

plugins/src/gcc/imp.rs
thiblahute: Indeed, let me retimestamp buffers at this point.

thiblahute: Right I got confused, let me remove pts from buffers individually

Also I did verify that we were not waiting on the clock and we weren't (because we sync on raw buffers already iirc)

plugins/src/gcc/imp.rs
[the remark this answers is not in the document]
thiblahute: We are talking about 5ms budgets here.

plugins/src/webrtcsink/imp.rs
thiblahute: Moved it into its own file (in a dedicated commit)

plugins/src/gcc/imp.rs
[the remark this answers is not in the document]
thiblahute: OK, done.

plugins/src/gcc/imp.rs
```diff
@@ -0,0 +1,1360 @@
...
+    max_bitrate: Bitrate,
+
+    detector: Detector,
+
+    clock_entry: Option<gst::SingleShotClockId>,
+
+    // Implemented like a leaky bucket
+    buffers: VecDeque<gst::Buffer>,
+    // Number of bits over the budget sent during last burst
+    last_overshot_bits: u64,
```
MathieuDuponchelle, line 726: fairly sure making this an `i64` would simplify the code ;)

plugins/src/gcc/imp.rs
MathieuDuponchelle: right, my point was that if it is implemented, then shouldn't we do what the comment says and remove it? Sorry for replying only now, but you marked the discussion as resolved ..

plugins/src/gcc/imp.rs
MathieuDuponchelle: @thiblahute ?

plugins/src/gcc/imp.rs
thiblahute: The comment was removed as it wasn't making sense, which is why the comment was discussed is resolved.

plugins/src/gcc/imp.rs
thiblahute: Look at the current code?

``` rs
    fn push_list(&self, list: gst::BufferList) {
        let res = self.srcpad.push_list(list);

        self.state.lock().unwrap().flow_return = res;
    }
```

this was solved for a good reason.

plugins/src/gcc/imp.rs
thiblahute: It should **never** be negative.

plugins/src/gcc/imp.rs
MathieuDuponchelle: What do you mean the comment was removed? It is clearly still part of the merge request, or I'm going blind

plugins/src/gcc/imp.rs
MathieuDuponchelle: what I'm saying is you could reverse the meaning of the field, add it to the budget instead of remove it, and the code will be cleaner looking ..

plugins/src/gcc/imp.rs
MathieuDuponchelle: It wasn't solved @thiblahute !

MathieuDuponchelle: also you're now assigning "void" to `_`. Can you explain why you don't directly post error messages instead of relying on upstream failing to chain to error out ooc?

plugins/src/gcc/imp.rs
thiblahute: I am blind :/ I must have expanded the wrong `mod.rs` in the github UI. Sorry about that.

I reworked the comment now so that we keep the behavior in any case, it shouldn't happen anymore when stuffing is on, but still.

plugins/src/gcc/imp.rs
thiblahute: Sensibly, but I can't find a good name for it, in practice that is the change I could make:

``` diff
diff --git a/plugins/src/gcc/imp.rs b/plugins/src/gcc/imp.rs
index 3b3bbd3..69c775f 100644
--- a/plugins/src/gcc/imp.rs
+++ b/plugins/src/gcc/imp.rs
@@ -722,8 +722,8 @@ struct State {
 
     // Implemented like a leaky bucket
     buffers: VecDeque<gst::Buffer>,
-    // Number of bits over the budget sent during last burst
-    last_overshot_bits: u64,
+    // Number of bits over the budget sent during last burst, always <= 0
+    last_overshot_bits: i64,
 
     flow_return: Result<gst::FlowSuccess, gst::FlowError>,
     last_push: time::Instant,
@@ -763,7 +763,7 @@ impl State {
                 self.estimated_bitrate as i64,
                 gst::ClockTime::SECOND.nseconds() as i64,
             )
-            .unwrap() - self.last_overshot_bits as i64;
+            .unwrap() + self.last_overshot_bits;
         let total_budget = budget;
         let mut remaining = self.buffers.iter().map(|b| b.size() as f64).sum::<f64>() * 8.;
         let total_size = remaining;
@@ -799,7 +799,7 @@ impl State {
         );
 
         self.last_push = now;
-        self.last_overshot_bits = if budget < 0 && !leaked { -budget as u64 } else { 0 };
+        self.last_overshot_bits = if !leaked { budget } else { 0 };
 
         list
     }
 ```

I don't think it is worth it.

plugins/src/gcc/imp.rs
thiblahute: WTF github hides unsolved conversations... oO

I took example on gstqueue to let upstream handle the flow error, but I didn't notice that queue also pauses the task when that happens. It is now done that way, meaning that `self.push_list` now returns the value and in the GstTask callback I pause and return when downstream returns an error.

MathieuDuponchelle:
Well, that was a trip, but we made it :)

PR #73, opened 2022-07-07
Allow passing free form metadata about consumers/producers/listenners

thiblahute:
In some applications people might need to be able to pass more metadata
than the only "display-name" this commit allow that by adding a field in
all signalling structure where it is free form json.

On the "webrtcsink" side this commit adds a `meta` property as a
gst::Structure which will be passed as a json string to the signalling
server so we enforce data to be structured.

### DIFF
--- a/plugins/src/signaller/imp.rs
+++ b/plugins/src/signaller/imp.rs
@@ -4,10 +4,12 @@ use async_std::task;
 use async_tungstenite::tungstenite::Message as WsMessage;
 use futures::channel::mpsc;
 use futures::prelude::*;
-use gst::glib;
 use gst::glib::prelude::*;
+use gst::glib::{self, Type};
+use gst::prelude::*;
 use gst::subclass::prelude::*;
 use once_cell::sync::Lazy;
+use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::Mutex;
 use webrtcsink_protocol as p;
@@ -98,10 +100,14 @@ impl Signaller {
             Ok::<(), Error>(())
         });
 
+        let meta = if let Some(meta) = element.property::<Option<gst::Structure>>("meta") {
+            serialize_value(&meta.to_value())
+        } else {
+            None
+        };
+
         websocket_sender
-            .send(p::IncomingMessage::Register(p::RegisterMessage::Producer {
-                display_name: element.property("display-name"),
-            }))
+            .send(p::IncomingMessage::Register(p::RegisterMessage::Producer { meta, }))
             .await?;
 
         let element_clone = element.downgrade();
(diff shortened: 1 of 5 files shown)


joaoantoniocardoso:
Nice, this will be useful to me, it might be worth mentioning this feature somewhere in the documentation, what do you think? :)

thiblahute:
It will be in `gst-inspect` and once we finally generate documentation for plugins written in rust it should appear there, not sure where else I could add that.

joaoantoniocardoso:
> It will be in `gst-inspect` and once we finally generate documentation for plugins written in rust it should appear there, not sure where else I could add that.

Oh, sorry, I didn't realize that, I think that is enough =)

thiblahute:
> > Sounds like you changed your mind then :-) - Happy to do that !
> 
> I don't think I did, but if we go for the more generic arbitrary metadata, the more specific one can be discarded, hopefully the property hadn't existed for long enough for someone to miss it :P

Right, I think I proposed that when we discussed it and I understood you prefered to keep it (while I thought removing it was better), anyway we don't care... I removed it now ;)

thiblahute:
ping @MathieuDuponchelle

www/webrtc.js
```diff
@@ -342,16 +342,21 @@ function session_closed(peer_id) {
...
+    console.log("Meta: ", JSON.stringify(meta));
 
     var nav_ul = document.getElementById("camera-list");
 
-    if (display_name == null) {
-        var li_str = '<li id="peer-' + peer_id + '"><button class="button button1">' + peer_id + '</button></li>';
-    } else {
-        var li_str = '<li id="peer-' + peer_id + '"><button class="button button1">' + display_name + '</button></li>';
+    meta = meta ? meta : {"display-name": peer_id};
+    let display_html = `${meta["display-name"] ? meta["display-name"] : peer_id}<ul>`;
```
MathieuDuponchelle, line 351: oh fancy, had no idea you could do templates with backticks in js

www/webrtc.js
```diff
@@ -393,10 +398,12 @@ function onServerMessage(event) {
     } else if (msg.type == "list") {
         clearPeers();
         for (i = 0; i < msg.producers.length; i++) {
-            addPeer(msg.producers[i].id, msg.producers[i].displayName);
+            console.error(`Msg: ${JSON.stringify(msg)}`)
```
MathieuDuponchelle, line 401: that doesn't seem needed

www/webrtc.js
```diff
@@ -393,10 +398,12 @@ function onServerMessage(event) {
...
+            addPeer(msg.producers[i].id, msg.producers[i].meta);
         }
     } else if (msg.type == "producerAdded") {
-        addPeer(msg.peerId, msg.displayName);
+        console.error(`Msg: ${JSON.stringify(msg)}`)
```
MathieuDuponchelle, line 405: that neither

plugins/src/signaller/imp.rs
```diff
@@ -418,3 +424,48 @@ impl ObjectImpl for Signaller {
...
+                    values
+                        .iter()
+                        .map(|value| value.nick())
+                        .collect::<Vec<&str>>()
+                        .join("+")
+                        .into(),
+                )
+            } else if let Ok(value) = val.serialize() {
+                Some(value.as_str().into())
+            } else {
```
MathieuDuponchelle, line 466: I think we want to log here

MathieuDuponchelle:
OK for me otherwise :+1:

MathieuDuponchelle:
@compiaffe this might be relevant for you (`display-name` removal)

compiaffe:
Thanks for the ping.
FYI @Hejsil @KunalAivero

MathieuDuponchelle:
Thanks!

PR #77, opened 2022-07-15
webrtcsink: Add AV1 support

yatinmaan:
This depends on [rtpav1](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/802/) and [some changes to av1enc](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2743) so those would need to be merged first.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -244,10 +244,16 @@ pub struct WebRTCSink {
 impl Default for Settings {
     fn default() -> Self {
         Self {
-            video_caps: ["video/x-vp8", "video/x-h264", "video/x-vp9", "video/x-h265"]
-                .iter()
-                .map(|s| gst::Structure::new_empty(s))
-                .collect::<gst::Caps>(),
+            video_caps: [
+                "video/x-vp8",
+                "video/x-h264",
+                "video/x-vp9",
+                "video/x-av1",
+                "video/x-h265",
+            ]
+            .iter()
+            .map(|s| gst::Structure::new_empty(s))
+            .collect::<gst::Caps>(),
             audio_caps: ["audio/x-opus"]
                 .iter()
                 .map(|s| gst::Structure::new_empty(s))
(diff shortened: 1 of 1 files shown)


plugins/src/webrtcsink/imp.rs
```diff
@@ -520,14 +546,25 @@ fn setup_encoding(
...
+                // Quirk: nvh264enc can perform conversion from RGB formats, but
+                // doesn't advertise / negotiate colorimetry correctly, leading
+                // to incorrect color display in Chrome (but interestingly not in
+                // Firefox). In any case, restrict to exclude RGB formats altogether,
+                // and let videoconvert do the conversion properly if needed.
+                structure_builder = structure_builder
+                    .field("format", &gst::List::new(&[&"NV12", &"YV12", &"I420"]));
+            }
+            "rav1enc" => {
+                // Unfortunately other formats don't yet work.
```
sdroege, line 560: Why not?

MathieuDuponchelle: Yes, please provide more information here, the existing restriction is a workaround, which is why I tried to document it as precisely as possible :)

MathieuDuponchelle:
looks good to me otherwise

plugins/src/webrtcsink/imp.rs
yatinmaan: Presumably, due to buggy encoder elements.

By default `rav1enc` would negotiate `Y444_12LE` and segfault.
`av1enc` (and `rav1enc` for other less verbose formats) seemed to work fine but Chromium fails to decode any of the frames sent over.

I could only get `GRAY8`, `I420` and `YV12` to actually render.

sdroege: Interesting, that should probably be reported. If I understand you correctly, `av1enc` with `Y444_12LE` was working fine with chrome?

Can you create issues against rav1e (not the GStreamer plugin) for that and CC me there?

yatinmaan: > If I understand you correctly, av1enc with Y444_12LE was working fine with chrome?

No, the `av1enc` element doesn't have caps for `Y444_12LE`. For `Y444`, its same as other formats where chrome receives the frames but fails to decode them.

> Can you create issues against rav1e (not the GStreamer plugin) for that and CC me there?

The segfault was from sometime ago. I rebuilt it now and that's fixed already. `Y444_12LE` just fails to render in chrome like the others now.

Interestingly `gst-launch-1.0 videotestsrc ! video/x-raw,height=480,width=640,framerate=30/1,format="Y42B" ! av1enc ! rsdav1ddec ! videoconvert ! autovideosink` fails for me. But it works if I use `av1dec` instead of `dav1d`.

sdroege: In any case, please create issues about all this and link them from the code here with some more explanation :)

yatinmaan: Chrome only supports the `main` profile for AV1 at the moment.

The issue here is that `av1enc` doesn't set its sink caps based on downstream `profile` [1] (even when its present). Also, we need to interpret a missing `profile` parameter in the SDP to mean to the `main` profile. We would probably need to special case this optional/missing param flow even after the `av1enc` bug is fixed.

A similar problem crashes Firefox for `vp9enc` [2]. Its decoder only supports the `main` profile for VP9.

For now, I updated the code to set the appropriate `raw_filter` based on downstream `profile[-id]` and added comments for the respective issues.

[1] https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1404
[2] https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1403

MathieuDuponchelle:
What's the status here? :)

sdroege:
I think this is blocked by https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/884#note_1569937 at least

MathieuDuponchelle:
> I think this is blocked by https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/884#note_1569937 at least

OK thanks, would be nice to see this go forward :)

MathieuDuponchelle:
Closing @yatinmaan please re-open against https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs when ready :)

sdroege:
I think it would be ready now

MathieuDuponchelle:
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/2743 is still listed as a dependency at the top though?

PR #85, opened 2022-08-19
Fix max-bitrate property setter

gondaruk:
`max-bitrate` was interpreted incorrectly when specified

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -2380,7 +2380,7 @@ impl ObjectImpl for WebRTCSink {
                 settings.cc_info.max_bitrate = (value.get::<u32>().expect("type checked upstream")
                     as f32
                     * if settings.do_fec {
-                        settings.cc_info.max_bitrate as f32 * 1.5
+                        1.5
                     } else {
                         1.
                     }) as u32;


gondaruk:
Hey,
sorry, i have not found any contribution guide so,
please, let me know whether I need to do anything else, to get this merged.

Thanks!

MathieuDuponchelle:
oof yes, @thiblahute care to validate this?

MathieuDuponchelle:
to be honest, this is very weird, the taking of fec into account should only happen at the point of usage of the property value, and shouldn't affect the value we store directly, can you correct this @thiblahute ?

MathieuDuponchelle:
for instance, this obviously breaks if do-fec is modified afterwards :)

thiblahute:
> to be honest, this is very weird, the taking of fec into account should only happen at the point of usage of the property value, and shouldn't affect the value we store directly, can you correct this @thiblahute ?

Yeah that happened when I moved to a dedicated structure for cc_info and during dev I moved from changing the semantics so that max-bitrate was actually the maximum bitrate and then back to your semantics.... but I still don't like the current semantics tbh..


See https://github.com/centricular/webrtcsink/pull/85#issuecomment-1221073873

gondaruk:
Thanks, @MathieuDuponchelle @thiblahute!
Closing this in favour of #86.

PR #86, opened 2022-08-19
webrtcsink: Fix the way we handle max-bitrate

thiblahute:
Computation of actual max bitrate was broken and in the end it is simpler
to keep the value set by the user and take into account the fec only
when required.

See https://github.com/centricular/webrtcsink/pull/85#issuecomment-1221073873


gondaruk:
Tested and it works like charm in my case, thanks.

MathieuDuponchelle:
thanks :)

PR #87, opened 2022-08-22
gcc: Take fec and number of encoders into account

gondaruk:
Fix case, when `bitrate` goes slightly higher than `max-bitrate` which leads to `fec-percentage` being out of range.

### DIFF
--- a/plugins/src/webrtcsink/imp.rs
+++ b/plugins/src/webrtcsink/imp.rs
@@ -48,6 +48,9 @@ const DEFAULT_DO_FEC: bool = true;
 const DEFAULT_DO_RETRANSMISSION: bool = true;
 const DEFAULT_ENABLE_DATA_CHANNEL_NAVIGATION: bool = false;
 const DEFAULT_START_BITRATE: u32 = 2048000;
+/* Start adding some FEC when the bitrate > 2Mbps as we found experimentally
+ * that it is not worth it below that threshold */
+const DO_FEC_THRESHOLD: u32 = 2000000;
 
 #[derive(Debug, Clone, Copy)]
 struct CCInfo {
@@ -139,8 +142,12 @@ struct Session {
     webrtc_pads: HashMap<u32, WebRTCPad>,
     peer_id: String,
     encoders: Vec<VideoEncoder>,
-    // Our Homegrown controller
+
+    // Our Homegrown controller (if cc_info.heuristic == Homegrown)
     congestion_controller: Option<CongestionController>,
+    // Our BandwidthEstimator (if cc_info.heuristic == GoogleCongestionControl)
+    rtpgccbwe: Option<gst::Element>,
+
     sdp: Option<gst_sdp::SDPMessage>,
     stats: gst::Structure,
 
@@ -732,6 +739,7 @@ impl Session {
         webrtcbin: gst::Element,
         peer_id: String,
         congestion_controller: Option<CongestionController>,
+        rtpgccbwe: Option<gst::Element>,
         cc_info: CCInfo,
     ) -> Self {
         Self {
@@ -742,6 +750,7 @@ impl Session {
             cc_info,
             rtprtxsend: None,
             congestion_controller,
+            rtpgccbwe,
             stats: gst::Structure::new_empty("application/x-webrtc-stats"),
             sdp: None,
             webrtc_pads: HashMap::new(),
@@ -961,6 +970,11 @@ impl Session {
             }
 
             self.encoders.push(enc);
+
+            if let Some(ref rtpgccbwe) = self.rtpgccbwe.as_ref() {
+                let max_bitrate = self.cc_info.max_bitrate * (self.encoders.len() as u32);
+                rtpgccbwe.set_property("max-bitrate", max_bitrate);
+            }
         }
 
         let appsrc = appsrc.downcast::<gst_app::AppSrc>().unwrap();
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
hrm, isn't that bitrate clamped properly at this point? If not, shouldn't we fix that instead?  :)

gondaruk:
Looks like before gcc was introduced, `target_bitrate` was used in calculation which was [always](https://github.com/centricular/webrtcsink/blob/64f664c859af180679bebc10ed44ee78793d54f5/plugins/src/webrtcsink/imp.rs#L1024) <= max_bitrate.
Here, looks like actual `bitrate` is used, which can be more than `max_bitrate`.

gondaruk:
Hmm. My bad. `bitrate` in this context is estimated bitrate notified by `rtpgccbwe`. Maybe, you're right and `rtpgccbwe` should not estimate more than `max_bitrate`?

gondaruk:
@MathieuDuponchelle sorry for confusion. I should have check deeper before doing PR.

Here is what I found.
If `do_fec` is set to true - `rtpgccbwe`'s `max-bitrate` [will be](https://github.com/centricular/webrtcsink/blob/main/plugins/src/webrtcsink/imp.rs#L1367) `max-bitrate * 1.5`.

So, `bitrate` in this callback is clamped but with different "max", because it includes the bitrate which is supposed to be used for fec.
I've updated the PR to contain `fec` check as well.

Let me know what you think. Thanks :)

MathieuDuponchelle:
@thiblahute ? this is all a bit ugly, not sure how to make nicer :)

thiblahute:
I think the way to make it nicer would be to have the `max-bitrate` take FEC into account all the time and then set the encoder bitrate / fec porcentage as needed, what do you think?

MathieuDuponchelle:
> I think the way to make it nicer would be to have the `max-bitrate` take FEC into account all the time and then set the encoder bitrate / fec porcentage as needed, what do you think?

You mean that it should also be the meaning of the prop as exposed to the user? I suppose that's a solution

MathieuDuponchelle:
@gondaruk , I guess that needs a decision, I say we go with what thibault suggests if you agree :)

gondaruk:
Thanks!
Let me work on that and I'll get back :)

gondaruk:
Hey @MathieuDuponchelle @thiblahute,
whenever you have a chance, could you please take a look on the update and let me know if it's what you meant or if I misunderstood you?
Thanks!

MathieuDuponchelle:
care to review this one @thiblahute ? :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -1749,29 +1744,28 @@ impl WebRTCSink {
         let mut state = element.imp().state.lock().unwrap();
 
         if let Some(consumer) = state.consumers.get_mut(peer_id) {
+            // Start adding some FEC when the bitrate > 2Mbps as we found experimentally
```
thiblahute, line 1747: Maybe add that comment around the constant definition?

plugins/src/webrtcsink/imp.rs
```diff
@@ -1749,29 +1744,28 @@ impl WebRTCSink {
...
+            // that it is not worth it below that threshold
             let fec_ratio = {
-                // Start adding some FEC when the bitrate > 2Mbps as we found experimentally
-                // that it is not worth it below that threshold
-                if bitrate <= 2_000_000 || consumer.cc_info.max_bitrate <= 2_000_000 {
+                if bitrate <= DO_FEC_THRESHOLD || consumer.cc_info.max_bitrate <= DO_FEC_THRESHOLD {
```
thiblahute, line 1750: `consumer.cc_info.max_bitrate <= DO_FEC_THRESHOLD` should be unnecessary as bitrate can't be > max_bitrate?

plugins/src/webrtcsink/imp.rs
```diff
@@ -1749,29 +1744,28 @@ impl WebRTCSink {
...
-                    (bitrate as f64 - 2_000_000.)
-                        / (consumer.cc_info.max_bitrate as f64 - 2_000_000.)
+                    (bitrate as f64 - DO_FEC_THRESHOLD as f64)
+                        / (consumer.cc_info.max_bitrate as f64 - DO_FEC_THRESHOLD as f64)
                 }
             };
 
-            let fec_percentage = fec_ratio * 50f64;
-            let encoders_bitrate = ((bitrate as f64) / (1. + (fec_percentage / 100.))) as i32;
+            let fec_percentage = (fec_ratio * 50f64) as u32;
```
thiblahute, line 1758: My understanding is that `bitrate` is the full bitrate for a particular stream, including the FEC percentage, in practice this means that we need to set the `encoder_bitrate` as bitrate / fec_percentage as it was the case before.'

Why do you think it is not the case anymore?

(we should add unit tests :D)

plugins/src/webrtcsink/imp.rs
```diff
@@ -1749,29 +1744,28 @@ impl WebRTCSink {
...
 
             if let Some(ref rtpxsend) = consumer.rtprtxsend.as_ref() {
                 rtpxsend.set_property("stuffing-kbps", (bitrate as f64 / 1000.) as i32);
             }
 
             for encoder in consumer.encoders.iter_mut() {
-                encoder.set_bitrate(element, encoders_bitrate);
+                encoder.set_bitrate(element, bitrate as i32);
```
thiblahute, line 1765: This reminds me that here we do not take into account the number of encoders. I think that  `webrtcsink::max-bitrate` should be the maximal bitrate **per stream**, meaning that when we set it to the bandwidth estimator we should do `max-bitrate = webrtcsink.max_bitrate * consumer.encoders.len()` and here we should end up with:

```
encoder_bitrate = ((((bitrate as f64) / (1. + (fec_percentage / 100.)))) / consumer.encoders.len()) as i32
```

@MathieuDuponchelle do you agree?

plugins/src/webrtcsink/imp.rs
gondaruk: Makes sense. Will do :)

plugins/src/webrtcsink/imp.rs
gondaruk: Makes sense. I've also noticed that `do_fec` wasn't considered here, so part of bitrate for fec was always substracted from the encoders bitrate (even if fec is disabled).

plugins/src/webrtcsink/imp.rs
gondaruk: I tried to do that, however, at the moment, when `rtpgccbwe` is created, `session.encoders` vector is still empty. I'd appreciate, if you could advise how/if this should be solved.

plugins/src/webrtcsink/imp.rs
gondaruk: I don't :) 
You're right, fixed that. Thank you!

plugins/src/webrtcsink/imp.rs
thiblahute: I think we need to keep a ref to the bandwith estimating element and set its `max-bitrate` each time an encoder is added.

gondaruk:
@thiblahute @MathieuDuponchelle, whenever you have a chance, could you please take a look?
Thanks :)

plugins/src/webrtcsink/imp.rs
```diff
@@ -1362,48 +1376,49 @@ impl WebRTCSink {
...
-                                return None;
-                            },
-                            Ok(e) => {
-                                let max_bitrate = (settings.cc_info.max_bitrate as f64 * if settings.do_fec { 1.5 } else { 1.}) as u32;
-                                e.set_properties(&[
+                let rtpgccbwe = match gst::ElementFactory::make("rtpgccbwe", None) {
+                    Err(err) => {
+                        glib::g_warning!(
+                            "webrtcsink",
+                            "The `rtpgccbwe` element is not available \
```
thiblahute, line 1385: Indentation looks fishy, did you run `cargo fmt`?

plugins/src/webrtcsink/imp.rs
```diff
@@ -1362,48 +1376,49 @@ impl WebRTCSink {
...
+                                            not doing any congestion control: {err:?}"
+                        );
+                        None
+                    }
+                    Ok(cc) => {
+                        webrtcbin.connect_closure(
+                            "request-aux-sender",
+                            false,
+                            glib::closure!(@watch element, @strong session_id, @weak-allow-none cc
```
thiblahute, line 1394:
```suggestion
                            glib::closure!(@watch element, @strong session_id, @watch cc
```

Maybe?

thiblahute (commented):
This looks good to me.

plugins/src/webrtcsink/imp.rs
gondaruk: Yes. However, looks like `cargo fmt` don't mind if I manually set second line indentation.
First argument `"webrtcsink"` though is always moved to the new line. Not sure why.

Feel free to reopen the conversation if updated version doesn't look good to you ;)

plugins/src/webrtcsink/imp.rs
gondaruk: So am I :) 
Thanks for review!

plugins/src/webrtcsink/imp.rs
gondaruk: Thanks for suggestion. 
This was basically done. Feel free to reopen the conversation, if needed :)

gondaruk:
Updated PR heading to reflect what was implemented at the end.

Let me know, if I can do anything else to get this merged. Thanks :)

gondaruk:
@MathieuDuponchelle, kindly reminder about this PR :)

MathieuDuponchelle:
OK let's go then :)

MathieuDuponchelle:
Thanks again @gondaruk :)

Issue #90, opened 2022-08-30
Deadlock on shutdown with gcc as the congestion controller


MathieuDuponchelle:
Reproduce with: `gst-launch-1.0 videotestsrc ! webrtcsink`, start a consumer then ctrl + C. Instead of shutting down gracefully, gst-launch stays stuck. Setting any other congestion control algo resolves the issue.

Copy paste from gdb run: https://gist.github.com/MathieuDuponchelle/4a4894078b0f62bb948b8cd3370a0cbe

Thread 40 is interesting:

```
https://gist.github.com/MathieuDuponchelle/4a4894078b0f62bb948b8cd3370a0cbe
```

cc @thiblahute :)

PR #91, opened 2022-08-31
gcc: Fix deadlock on shutdown

thiblahute:
We were holding the element .state lock while trying to shutdown leading
to a deadlock.

Fixes: https://github.com/centricular/webrtcsink/issues/90


MathieuDuponchelle:
Great, thanks :)

PR #93, opened 2022-09-05
Fix consumer disconnection

### LINKED ISSUE #92: Server doesn't fully end a broken connection
If the client is unexpectedly disconnected from the signalling server, the session on the server will remain open until the stream is finished.
This can be especially problematic on low-powered devices, which could have significant performance issues when wasting processing power and memory.

Steps to reproduce:

1. Run the demo
2. Add many (roughly 10 to 20 streams) to see a noticeable difference in processor usage
3. Close the browser window or refresh the page
4. Even after a long time, gst-launch will keep using resources
5. When the stream ends, many end_session function calls from webrtcsink_signalling::handlers can be seen in the debug terminal

My guess would be that the session isn't correctly ended in this case, however I do not have any Rust experience, so it's hard to pinpoint the exact source of the issue.

MathieuDuponchelle:
Fixes #92

### DIFF
--- a/signalling/src/handlers/mod.rs
+++ b/signalling/src/handlers/mod.rs
@@ -120,7 +120,7 @@ impl Handler {
             .sessions
             .iter()
             .filter_map(|(session_id, session)| {
-                if session.producer == peer_id {
+                if session.producer == peer_id || session.consumer == peer_id {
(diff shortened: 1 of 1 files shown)


thiblahute:
Thanks for looking at it!

PR #104, opened 2022-09-29
signalling: Do not stop consuming sessions when a peer stops being a …

thiblahute:
…producer

This is a regression from: https://github.com/centricular/webrtcsink/pull/93


MathieuDuponchelle:
Closing in favor of https://github.com/centricular/webrtcsink/pull/106

PR #106, opened 2022-09-30
signalling: track consumer and producer sessions in dedicated HashMaps

MathieuDuponchelle:
This avoids full traversals of the sessions HashMap.

Also fixes accidental session teardown when a producer is stopped while also in a session as a consumer.


thiblahute:
Lgtm

Issue #110, opened 2022-10-11
Fatal runtime error: failed to initiate panic, error 5. Aborted (core dumped) when closing the web


NanNguyen01:
Hi, 
I tried the latest main version of the code using the provided example run instructions. When I opened a browser's tab, I clicked on stream's peer-id about one or more times, then close the browser's tab, it would occasionally produce the following runtime error.
![image](https://user-images.githubusercontent.com/101386916/194988923-4de4c1fe-695a-4799-9bd7-32e0abc4130f.png)

MathieuDuponchelle:
cc @thiblahute :)

PR #112, opened 2022-10-11
gcc: Only log error when pausing task failes

thiblahute:
We can't do much about it at this point

Fixes https://github.com/centricular/webrtcsink/issues/110

### DIFF
--- a/plugins/src/gcc/imp.rs
+++ b/plugins/src/gcc/imp.rs
@@ -1054,7 +1054,7 @@ impl BandwidthEstimator {
         self.srcpad.start_task(move || {
             let pause = || {
                 if let Some(pad) = weak_pad.upgrade() {
-                    pad.pause_task().unwrap();
+                    let _ = pad.pause_task();
                 }
             };
             let bwe = weak_bwe


MathieuDuponchelle:
The only reason pause_task can return False is when the pad had no task apparently, how can this be happening?

MathieuDuponchelle:
Also I notice that we might pause while holding the state lock, that's not ideal (and something I fixed in mccparse / sccparse some time ago)

MathieuDuponchelle:
2131d5bfda964a062127ed859ce8d28e17f0e450 in gst-plugins-rs fwiw

MathieuDuponchelle:
scratch that last remark, I think we always release actually :)

thiblahute:
> The only reason pause_task can return False is when the pad had no task apparently, how can this be happening?

From the stack trace this happened when pushing the buffer list failed, I guess the task is already stopped from some other code path and is being joined?

MathieuDuponchelle:
I see, then let's just ignore altogether and not log an error :)

MathieuDuponchelle:
elsewhere in `gst-plugins-rs` we just `let _ = pause_task`, let's just do the same here :)

MathieuDuponchelle:
Thanks :)

